PR #2, opened 2018-08-21
Add readme badges

alexkaserbacher:


fdw:
So there's the sign-off missing on some commits (maybe https://git-scm.com/docs/git-config/1.8.2#git-config-formatsignoff would be a good idea?).
Second, instead of merging master into feature branches, could we please rebase onto master? I find the history a lot cleaner that way.

PR #9, opened 2018-08-23
Add patch to package.json

alexkaserbacher:
Add patch to the install-process of the frontend, such that the Swagger UI-dependencies work properly

Signed-off-by: Alexander Kaserbacher <alexander.kaserbacher@tngtech.com>


frontend/patch.js
```diff
@@ -0,0 +1,13 @@
+const fs = require('fs');
```
fdw: How about a short comment why this patch is needed?

PR #10, opened 2018-08-23
Extend specification-form to also accept .yaml files

alexkaserbacher:


frontend/src/app/specification-form/specification-form.component.html
```diff
@@ -13,7 +13,7 @@
   <div class="row top-buffer">
     <div class="col">
       <p>
-        <input type="file" placeholder="Upload Swagger Specification" accept=".yml,.json" ngModel
+        <input type="file" placeholder="Upload Swagger Specification" accept=".yml,.json,.yaml" ngModel
```
fdw: That's gonna be interesting once we have different spec formats \o/

PR #18, opened 2018-09-10
Add API descriptions.

alexkaserbacher:
Add descriptions to models and endpoints to make the generated documentation better readable.

### DIFF
--- a/backend/src/main/kotlin/com/tngtech/apicenter/backend/config/WebConfiguration.kt
+++ b/backend/src/main/kotlin/com/tngtech/apicenter/backend/config/WebConfiguration.kt
@@ -35,7 +35,6 @@ class WebConfiguration {
     private fun apiMetadata() = ApiInfoBuilder()
         .title("ApiCenter REST API")
         .description("REST API for ApiCenter - a repository for all your OpenAPI specifications")
-        .version("1.0.0")
         .license("Apache License Version 2.0, January 2004")
         .build()
 }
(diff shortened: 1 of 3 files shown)


backend/src/main/kotlin/com/tngtech/apicenter/backend/config/WebConfiguration.kt
```diff
@@ -29,4 +30,12 @@ class WebConfiguration {
         .apis(RequestHandlerSelectors.any())
         .paths(PathSelectors.any())
         .build()
+        .apiInfo(apiMetadata())
+
+    private fun apiMetadata() = ApiInfoBuilder()
+        .title("ApiCenter REST API")
+        .description("REST API for ApiCenter - a repository for all your OpenAPI specifications")
+        .version("1.0.0")
```
fdw: Can we save that version separately (maybe in den gradle.properties)?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/dto/SpecificationFileDto.kt
```diff
@@ -1,5 +1,13 @@
...
+import io.swagger.annotations.ApiModelProperty
 import java.util.UUID
 
-data class SpecificationFileDto @JsonCreator constructor(val fileContent: String?, val fileUrl: String? = "", val id: UUID? = null)
\ No newline at end of file
+@ApiModel(description = "Represents an OpenAPI-File")
+data class SpecificationFileDto @JsonCreator constructor(
+    @ApiModelProperty("The content of a file to be uploaded - either this or fileUrl has to be set.") val fileContent: String?,
+    @ApiModelProperty("The url of a remote file to be loaded - either this or fileContent has to be set.") val fileUrl: String? = "",
+    @ApiModelProperty("The uuid of an already existing specification (optional)") val id: UUID? = null
```
fdw: `@ApiModelProperty` has a property `required`, so that the "(optional)" is not needed.
What I'm wondering is whether we can somehow tell the spec that either `fileContent` or `fileUrl` is needed. If we make `SpecificationFileDto` an interface with two implementations, one for `fileContent` and one for `fileUrl`, do you think the deserialization will still work? That could be a bit cleaner.

PR #25, opened 2018-10-08
Specification versioning

alexkaserbacher:
This Pull Request achieves the following:

when you upload a new version of an existing specification (i.e. with the same name), it automatically creates a new version of the specification instead of a new specification.
Versions of specifications are shown in the frontend with an expandable table view.


backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
```diff
@@ -13,17 +14,32 @@ import javax.transaction.Transactional
...
 ) : SpecificationPersistenceService {
 
     override fun save(specification: Specification) {
-        specificationRepository.save(SpecificationEntity.fromDomain(specification))
+        val specificationEntity = specificationEntityMapper.fromDomain(specification)
+
+        if (specificationRepository.existsByTitle(specificationEntity.title)) {
+            val existingEntity = specificationRepository.findByTitle(specificationEntity.title)
+
+            val versions = existingEntity.versions + specificationEntity.versions
```
fdw: So if the new version already exists, it is overwritten? I'm not sure if we should prohibit that.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
```diff
@@ -13,17 +14,32 @@ import javax.transaction.Transactional
...
-    private val entityManager: EntityManager
+    private val entityManager: EntityManager,
+    private val specificationEntityMapper: SpecificationEntityMapper
 ) : SpecificationPersistenceService {
 
     override fun save(specification: Specification) {
-        specificationRepository.save(SpecificationEntity.fromDomain(specification))
+        val specificationEntity = specificationEntityMapper.fromDomain(specification)
+
+        if (specificationRepository.existsByTitle(specificationEntity.title)) {
```
fdw: Ideally, we could use a UUID for this instead of the name.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
```diff
@@ -13,17 +14,32 @@ import javax.transaction.Transactional
...
+            val existingEntity = specificationRepository.findByTitle(specificationEntity.title)
+
+            val versions = existingEntity.versions + specificationEntity.versions
+            val newEntity = SpecificationEntity(existingEntity.id, existingEntity.title, existingEntity.description, versions, existingEntity.remoteAddress)
+            newEntity.versions.map { versionEntity -> versionEntity.specification = newEntity }
+
+            specificationRepository.save(newEntity)
```
fdw: Maybe the whole block should be a new method?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionController.kt
```diff
@@ -0,0 +1,24 @@
...
+import com.tngtech.apicenter.backend.domain.handler.VersionHandler
+import org.springframework.web.bind.annotation.DeleteMapping
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import java.util.UUID
+
+@RestController
+@RequestMapping("/versions")
```
fdw: Shouldn't this rather be a subresource to `/specification`?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/SpecificationEntity.kt
```diff
@@ -1,44 +1,25 @@
...
 data class SpecificationEntity(
     @Id @GeneratedValue val id: UUID,
     @Field val title: String,
     @Field @Column(columnDefinition = "TEXT") val description: String?,
-    @Field val version: String,
-    @Field @Column(columnDefinition = "TEXT") val content: String,
+    @IndexedEmbedded @OneToMany(
+        mappedBy = "specification",
+        cascade = [CascadeType.ALL]
+    ) val versions: List<VersionEntity>,
```
fdw: Do we know that they're sorted?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/VersionEntity.kt
```diff
@@ -0,0 +1,20 @@
...
+import javax.persistence.Column
+import javax.persistence.Entity
+import javax.persistence.GeneratedValue
+import javax.persistence.Id
+import javax.persistence.ManyToOne
+
+@Entity
+@Indexed
+data class VersionEntity(
+    @Id @GeneratedValue val id: UUID,
```
fdw: Why does it need a UUID? It should be uniquely identifiable by spec and version.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
alexkaserbacher: What would your expected response be?
- Version is overwritten
- Version exists twice
- Return an error message?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
alexkaserbacher: But when you create a new Specification and want to check if some Specification with the same name already exists? That's the use case here.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
alexkaserbacher: What do you mean exactly?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/SpecificationEntity.kt
alexkaserbacher: They are implicitly sorted by creation date (which is exactly what we want).

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/VersionEntity.kt
alexkaserbacher: Because we need to address Versions directly from the frontend and I found unique IDs to come in handy here.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/SpecificationEntity.kt
alexkaserbacher: I will make the sorting explicit now.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
fdw: Either an outright error, or overwriting is permitted via a configuration option.

My main point is that you don't want to overwrite specs that other people depend on - just like Maven's Nexus. So, ideally, final versions cannot be changed (or only if the admin explicitely allows this), but SNAPSHOTs can. Since we do not yet have a concept of SNAPSHOTs, I'd prefer no overwriting ever.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
fdw: I'm not sure if we should prevent two specs from being called by the same name. There may be a legitimate reason for that.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
fdw: I think that the method does too much and would rather extract the stuff from the `if`-block into a new method.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
alexkaserbacher: I think we should go with the outright error then.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/VersionEntity.kt
fdw: Hm. I'm not sure if this isn't more confusing, since we now have two things that are addressable by UUIDs. Maybe a simple concatenation of Spec-UUID + Version would work?

alexkaserbacher: How about a query like `GET /specifications/{specificationId}?version={version}` 
or `GET /specifications/{specificationId}/{version}`
?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/VersionEntity.kt
```diff
@@ -3,18 +3,33 @@ package com.tngtech.apicenter.backend.connector.database.entity
...
+    ]
+)
 data class VersionEntity(
     @Id @GeneratedValue val id: UUID,
     @ContainedIn @ManyToOne var specification: SpecificationEntity?,
     @Field val version: String,
-    @Field @Column(columnDefinition = "TEXT") val content: String
-)
\ No newline at end of file
+    @Field @Column(columnDefinition = "TEXT") val content: String,
```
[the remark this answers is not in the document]
alexkaserbacher: Why should the content of a specification be a `DATE`? ;)

PR #27, opened 2018-10-10
User management

alexkaserbacher:
- Adds Atlassian Crowd as authentication provider to ApiCenter
- Adds support for JWT to authenticate user requests
- Adds frontend functionality to login and logout users


backend/src/main/kotlin/com/tngtech/apicenter/backend/config/SecurityCrowdConfiguration.kt
```diff
@@ -0,0 +1,97 @@
...
+    WebSecurityConfigurerAdapter() {
+
+    @Value("\${jwt.secret}")
+    private lateinit var jwtSecuritySecret: String;
+
+    override fun configure(httpSecurity: HttpSecurity) {
+        httpSecurity
+            .cors()
+            .and()
+            .csrf().disable()
```
fdw: Why?

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/SecurityCrowdConfiguration.kt
```diff
@@ -0,0 +1,97 @@
...
+
+@EnableWebSecurity
+@Configuration
+class SecurityCrowdConfiguration @Autowired constructor(private val userHandler: UserHandler) :
+    WebSecurityConfigurerAdapter() {
+
+    @Value("\${jwt.secret}")
+    private lateinit var jwtSecuritySecret: String;
+
+    override fun configure(httpSecurity: HttpSecurity) {
```
fdw: This method configures the ApiCenter itself, while the ones below only concern the configuration in regard to Crowd, right? Maybe split this file into two?

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/handler/SessionHandler.kt
```diff
@@ -0,0 +1,32 @@
...
+            searchUser
+        }
+
+        val userToStore = if (storedUser.email != searchUser.email || storedUser.username != searchUser.username) {
+            User(storedUser.id, searchUser.email, searchUser.username, "crowd", searchUser.externalId)
+        } else {
+            storedUser
+        }
+
+        if (userHandler.checkExistenceByOrigin(searchUser)) userHandler.store(userToStore)
```
fdw: Please use braces

backend/src/main/resources/api-definition.json
```diff
@@ -27,6 +27,16 @@
...
+            "in": "header",
+            "name": "Authorization",
+            "description": "JWT to authenticate the user, preceded with the string 'Bearer '",
+            "schema": {
+              "type": "string"
+            }
+          }
+        ],
         "responses": {
           "200": {
```
fdw: Now it should also be possible to get 403s, right?

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/SecurityCrowdConfiguration.kt
alexkaserbacher: Don't we want to allow requests from multiple sources (without the need to issue a CSRF token)? Client, CI-Server, Gradle Plugins etc.

backend/src/main/resources/api-definition.json
alexkaserbacher: Yes, it's already in there ;)

PR #28, opened 2018-10-10
Include mutation testing tools

alexkaserbacher:
- Split unit/integration tests in backend over different modules
- Include pitest for backend mutation testing
- Include stryker for frontend mutation testing
- Update readme


frontend/karma.conf.js
```diff
@@ -26,12 +26,6 @@ module.exports = function (config) {
     logLevel: config.LOG_INFO,
     autoWatch: true,
     browsers: ['Chrome'],
-    singleRun: false,
-    customLaunchers: {
```
fdw: Could you please rebase onto the latest master? I think we want to keep these changes.

alexkaserbacher: This is no issue with the history, I accidentally overwrote it - I will reintroduce the configuration to `karma.conf.js` manually.

PR #34, opened 2018-10-18
Remove GeneratedValue annotation from UUID in specification entity.

alexkaserbacher:
Remove GeneratedValue annotation in order to not unnecessarily generate the UUID a second time and causing the POST /specification endpoint to return the wrong id.


fdw:
I'm sorry if I'm just blind, but where is the UUID generated for the first time? If I understand this correctly, it has to be supplied by the client, as the REST entity already expects it (if you send content). Or what did I miss?

alexkaserbacher:
> I'm sorry if I'm just blind, but where is the UUID generated for the first time? If I understand this correctly, it has to be supplied by the client, as the REST entity already expects it (if you send content). Or what did I miss?

@fdw It happens within the mapping from DTO -> Domain. Look here: https://github.com/TNG/ApiCenter/blob/fix-spec-id-api/backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/mapper/converter/SpecificationConverter.kt#L33-L36

PR #35, opened 2018-12-08
Link to contributing from readme

boidolr:


fdw:
Unfortunately, we still need the DCO sign-off. Could you please add that?

Issue #21, opened 2018-09-19
Make stored specifications downloadable


alexkaserbacher:
Add a button that adds the possibility to download stored specifications.

PR #37, opened 2019-01-14
Makes stored specifications downloadable

catenoid:
In the specification overview list, two buttons are added to download the specification version as either JSON or YML.

Because the specifications are already available in browser memory in JSON format when a user views the specifcation overview list, no server interaction is required for this format. To download as YML, a request is sent to the server to perform JSON to YML conversion. The browser download starts when the button is clicked.

Currently the buttons look identical (both the OpenIcon "data-transfer-download", though they are distinguished by their hover text.

Issue #21


backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionController.kt
```diff
@@ -1,20 +1,47 @@
...
-    @GetMapping("/specifications/{specificationId}/versions/{version}")
-    fun findVersion(@PathVariable specificationId: UUID, @PathVariable version: String): VersionDto = versionDtoMapper.fromDomain(versionHandler.findOne(specificationId, version))
+    @RequestMapping("/specifications/{specificationId}/versions/{version}",
+        produces = ["application/json", "application/yml"],
+        headers = ["Accept=application/json", "Accept=application/yml"],
+        method = [RequestMethod.GET])
+    fun findVersion(@PathVariable specificationId: UUID, @PathVariable version: String, @RequestHeader("Accept") accept: String): VersionDto {
+        val foundVersion = versionHandler.findOne(specificationId, version)
+        if (accept == "application/yml") {
+            logger.info("YAML requested")
```
alexkaserbacher: Logging is a very good idea :) although, maybe the logging messages should be a little more verbose here.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionController.kt
```diff
@@ -1,20 +1,47 @@
...
+            val jsonNodeTree: JsonNode = ObjectMapper().readTree(foundVersion.content)
```
alexkaserbacher: Little hint: in Kotlin, you do not need to explicitly state the types, the compiler is able to infer them

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionController.kt
```diff
@@ -1,20 +1,47 @@
...
+            val jsonAsYaml: String = YAMLMapper().writeValueAsString(jsonNodeTree)
+            logger.info(jsonAsYaml)
+            return versionDtoMapper.fromDomain(Version(foundVersion.version, jsonAsYaml))
+        } else {
```
alexkaserbacher: We have two cases: yml or json. Here you explicitly handle _everything but yml_ as the default case. Maybe we could explicitly handle the case when json is uploaded and move some sort of error handling in the else-part.

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/ContentNegotiationConfiguration.kt
```diff
@@ -0,0 +1,23 @@
+package com.tngtech.apicenter.backend.config
+
+import com.fasterxml.jackson.databind.ObjectMapper
```
fdw: There are some imports that are not used

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/ContentNegotiationConfiguration.kt
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer
+import org.springframework.http.MediaType
+
+@Configuration
+class WebConfig : WebMvcConfigurer {
+
+  override fun configureContentNegotiation(configurer: ContentNegotiationConfigurer) {
+    configurer.
+            parameterName("mediaType").
+            ignoreAcceptHeader(true).
```
fdw: Why ignore the header? This is preferable to a query paramter, in my opinion.

frontend/src/app/specification-overview/specification-overview.component.ts
```diff
@@ -1,8 +1,10 @@
 import {Component, OnInit} from '@angular/core';
 import {SpecificationService} from '../specification.service';
 import {Specification} from '../models/specification';
+import {Version} from '../models/version';
 import {VersionService} from '../version.service';
 
+/* tslint:disable:member-ordering */
```
fdw: Why is this needed? Disabling the linter always looks like you're doing it wrong ;)

frontend/src/app/specification-overview/specification-overview.component.ts
```diff
@@ -38,6 +43,38 @@ export class SpecificationOverviewComponent implements OnInit {
     }
   }
 
+  public async downloadSpecification(specification: Specification, fileType: string) {
+    const latestVersion = specification.versions[0];
+    if (latestVersion !== undefined) {
```
fdw: If it is unspecified, what happens? And why could this ever happen?

frontend/src/app/specification-overview/specification-overview.component.ts
```diff
@@ -38,6 +43,38 @@ export class SpecificationOverviewComponent implements OnInit {
...
+      this.downloadVersion[fileType](specification, latestVersion);
+    }
+  }
+
+  private createDownloadFileName(specification: Specification, version: Version) {
+    // Make it filename safe by replacing any non-alphanumeric character with an underscore
+    return (specification.title + '_v' + version.version).replace(/[^a-z0-9\.]/gi, '_');
```
fdw: I'm not sure if I'd do this - they can change the file name; it is just a download after all. Or do you see a special case where this is important?

frontend/src/app/specification-overview/specification-overview.component.ts
```diff
@@ -38,6 +43,38 @@ export class SpecificationOverviewComponent implements OnInit {
...
   }
 
+  public async downloadSpecification(specification: Specification, fileType: string) {
+    const latestVersion = specification.versions[0];
+    if (latestVersion !== undefined) {
+      this.downloadVersion[fileType](specification, latestVersion);
+    }
+  }
+
+  private createDownloadFileName(specification: Specification, version: Version) {
```
fdw: Regarding ordering: I always like to structure code beginning with the most abstract methods and going deeper further down. Each method is beneath the one that calls it (if there's only one caller), or maybe beneath all of the callers.

frontend/src/app/specification-overview/specification-overview.component.ts
```diff
@@ -38,6 +43,38 @@ export class SpecificationOverviewComponent implements OnInit {
...
+  public downloadYAMLVersion = async (specification: Specification, version: Version) => {
+    const fileName = this.createDownloadFileName(specification, version);
+    this.versionService.getYAMLVersion(specification.id, version.version)
+      .subscribe(event => {
+        this.doDownload(event.content, fileName + '.yml', 'application/yaml');
+      });
+  }
+
+  public downloadVersion = {
+    // Avoids code duplication in this.downloadSpecification
```
fdw: That comment is not really needed, is it?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionController.kt
```diff
@@ -1,20 +1,50 @@
...
+
+private val logger = KotlinLogging.logger {}
 
 @RestController
 class VersionController constructor(private val versionHandler: VersionHandler, private val versionDtoMapper: VersionDtoMapper) {
 
-    @GetMapping("/specifications/{specificationId}/versions/{version}")
-    fun findVersion(@PathVariable specificationId: UUID, @PathVariable version: String): VersionDto = versionDtoMapper.fromDomain(versionHandler.findOne(specificationId, version))
+    @RequestMapping("/specifications/{specificationId}/versions/{version}",
+        produces = ["application/json", "application/yml"],
```
fdw: There should be constants for JSON, at least. Maybe you also can add one for yaml, somewhere?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionController.kt
```diff
@@ -1,20 +1,50 @@
...
+        headers = ["Accept=application/json", "Accept=application/yml"],
+        method = [RequestMethod.GET])
+    fun findVersion(@PathVariable specificationId: UUID,
+                    @PathVariable version: String,
+                    @RequestHeader(value="Accept", defaultValue=MediaType.APPLICATION_JSON_VALUE) accept: String = MediaType.APPLICATION_JSON_VALUE): VersionDto {
+        // i.e. The integration test and unit test require the default specified in two different ways
+        val foundVersion = versionHandler.findOne(specificationId, version)
+        if (accept == "application/yml") {
```
fdw: Instead of doing the content type negotiation by hand, I'd suggest using something like this: https://blog.jdriven.com/2016/11/handling-yaml-format-rest-spring-boot/ Let's add a new `ObjectMapper` for yaml and configure Spring Web MVC to use it as a possible `MessageConverter`. Then you don't have to do this in every method and it works as expected.

frontend/src/app/specification-overview/specification-overview.component.ts
catenoid: I suppose just "defensive programming"

fdw: Makes sense :) But I'd either expect an error message if it doesn't work, or we just assume that it has to work don't check.

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/ContentNegotiationConfiguration.kt
catenoid: This is required... when this line is deleted, there is a 406 error when requesting content other than JSON. Fraid I really can't explain why that happens.

frontend/src/app/specification-overview/specification-overview.component.ts
catenoid: Just thought it would be helpful, if you're downloading many at once.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionController.kt
catenoid: In the current configuration, if the user doesn't specify a file format, JSON is returned. So this made sense to me? Perhaps I haven't understood your comment 😅

.gitignore
```diff
@@ -38,6 +38,20 @@ crashlytics.properties
 crashlytics-build.properties
 fabric.properties
 
+# Database entities
+backend/com.tngtech.apicenter.backend.connector.database.entity.SpecificationEntity
```
fdw: None of these changes are needed after the other PR is merged, right?

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/ContentNegotiationConfiguration.kt
fdw: Okay, this is sensible after the discussion that the formats of the resource and of the content can be different. But maybe we should rename the parameter to something more inconspicuous like "downloadFormat".

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionController.kt
fdw: I'm not sure if it makes sense to do the conversion here or somewhere else. What do you say?

frontend/src/app/specification-overview/specification-overview.component.ts
fdw: Now I see what you mean :) The first part where the version is added is indeed useful. I was talking about the regexp.

frontend/src/app/specification-overview/specification-overview.component.ts
fdw: Obviously my last comment is lacking in the grammar department.
Correct: 
But I'd either expect an error message if it doesn't work, or we just assume that it has to work _and_ don't check.

frontend/src/app/specification-overview/specification-overview.component.ts
catenoid: Oh 😄 That's because I don't like spaces in my filenames...

.gitignore
catenoid: Correct that this change is also in the other PR. However that did not make these addtions to the gitignore here not useful.

Issue #36, opened 2019-01-09
Build a single deployable


fdw:
Ideally, you should only have to deploy one artifact instead of backend and frontend separately. Maybe the frontend can be included in the static resources of the backend?

PR #38, opened 2019-01-21
Config for a single deploy

catenoid:
Introduces an additional gradle task, which builds the frontend, and copies those generated Angular files into the "static resources" of the backend. The project can be run separately or combined (as before).

Copying the files is not enough: a user trying to access the homepage (ie. Angular's generated `index.html`, it's related `*.js` files etc.) gives a 403 error. This is because the backend would formerly only accept login attempts (path `/sessions`) from unauthenticated users. The only solution online is to individually permit all Angular's generated files.

Issue #36 

Useful:
- [Spring boot security with static resources](https://stackoverflow.com/questions/24916894/serving-static-web-resources-in-spring-boot-spring-security-application)
- [Single build with gradle and Angular](https://ordina-jworks.github.io/architecture/2018/10/12/spring-boot-angular-gradle.html)


backend/src/main/kotlin/com/tngtech/apicenter/backend/config/SecurityConfiguration.kt
```diff
@@ -25,12 +26,33 @@ class SecurityConfiguration : WebSecurityConfigurerAdapter() {
...
             .antMatchers("/sessions").permitAll()
-            .and()
-            .authorizeRequests()
+
+            .antMatchers("/").permitAll()
+            .antMatchers("/3rdpartylicenses.txt").permitAll()
+            .antMatchers("/favicon.ico").permitAll()
+            .antMatchers("/index.html").permitAll()
+            .antMatchers("/main.js").permitAll()
+                .antMatchers("/main.js.map").permitAll()
```
fdw: Could you please fix the indentation?

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/SecurityConfiguration.kt
```diff
@@ -25,12 +26,33 @@ class SecurityConfiguration : WebSecurityConfigurerAdapter() {
...
+            .antMatchers("/open-iconic.*").permitAll()
+            .antMatchers("/polyfills.js").permitAll()
+                .antMatchers("/polyfills.js.map").permitAll()
```
fdw: Do we really need to open the `map` files for everyone? I think they are just for developers, right?

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/SecurityConfiguration.kt
```diff
@@ -25,12 +26,33 @@ class SecurityConfiguration : WebSecurityConfigurerAdapter() {
...
+            .antMatchers("/runtime.js").permitAll()
+                .antMatchers("/runtime.js.map").permitAll()
+            .antMatchers("/styles.js").permitAll()
+                .antMatchers("/styles.js.map").permitAll()
+            .antMatchers("/scripts.js").permitAll()
+                .antMatchers("/scripts.js.map").permitAll()
+            .antMatchers("/vendor.js").permitAll()
+                .antMatchers("/vendor.js.map").permitAll()
             .anyRequest().authenticated()
```
fdw: What would happen if we did not define those expections?

frontend/package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "ng": "ng",
     "start": "ng serve",
-    "build": "ng build --prod",
+    "build": "ng build --prod=false",
```
fdw: Instead of hardcoding the dev environment, I'd suggest adding another task (here and in the backend's gradle): `gradle buildDev` and `gradle buildProd` for example which will build either the dev or the prod environment. What do you say?

fdw:
This could also be a good time to update to Gralde 5.1 ;)

fdw:
Oh, and some tests fail. Could you please have a look?

fdw:
One more thing: Can you please also update the Readme file with the new instructions?

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/SecurityConfiguration.kt
catenoid: Anyone trying to use ApiCenter's frontend eg. access the index page would get a 403 error, because the backend is not otherwise going to serve unauthorised users any static files.

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/SecurityConfiguration.kt
catenoid: This is partially true. I was guided by what console log messages indicated a resource script could not be loaded with a 403 response. Some of them are certainly needed, I listed all for safe measure, but can be a bit more thorough

frontend/package.json
catenoid: Interestingly, the build fails in when --prod is true, because the apiUrl is not set in the prod environment. Have I missed something?

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/SecurityConfiguration.kt
catenoid: Heh, there was me thinking this made more sense 😉 Sure.

catenoid:
All the tests I ran were sucessful: ng lint, ng test, ./gradlew test, ./gradlew integrationTest. Were the failures just because the build structure has changed?

gradlew.bat
```diff
@@ -1,84 +1,84 @@
-@if "%DEBUG%" == "" @echo off
```
catenoid: I think this is because git bash changed the Windows line endings.

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/SecurityConfiguration.kt
fdw: Are they then prompted to log in, or does it just fail?
If the exceptions are necessary, then I'd suggest to plainly allow `*.js*` and `*.js.map`

frontend/package.json
fdw: This is somewhat important. Can we set this to a sensible default value?

.gitignore
```diff
@@ -38,6 +38,20 @@ crashlytics.properties
 crashlytics-build.properties
 fabric.properties
 
+# Database entities
+backend/com.tngtech.apicenter.backend.connector.database.entity.SpecificationEntity
```
fdw: Do you know where this comes from?

.gitignore
```diff
@@ -38,6 +38,20 @@ crashlytics.properties
...
+
+# Single-deploy additional Gradle directories
+frontend/.gradle
```
fdw: How about `**/.gradle`?

.gitignore
```diff
@@ -38,6 +38,20 @@ crashlytics.properties
...
+.gradle/
+
+# Angular build directories
+frontend/build/
+backend/src/main/resources
```
fdw: I'm afraid this is too broad; sometimes, there are useful files in the folder. Can we just ignore the subfolder or something?

.gitignore
```diff
@@ -38,6 +38,20 @@ crashlytics.properties
...
+
+# Kotlin libraries
+lib/
+
 ### Intellij Patch ###
 # Comment Reason: https://github.com/joeblau/gitignore.io/issues/186#issuecomment-215987721
```
fdw: We can remove this

README.md
```diff
@@ -33,30 +33,48 @@ We plan to add OAuth 2 support shortly.
 
 ## Getting started
 
-ApiCenter consists of a RESTful backend service written in Kotlin and a SPA frontend in Angular. Both are contained in this git repo.
+ApiCenter consists of a RESTful backend service written in Kotlin and a SPA frontend in Angular. Both are contained in this git repo. There are two ways to run ApiCenter: one combines the tasks of the frontend and backend together, for simplicity; the other requires two terminal sessions, however has advantages for developers.
```
fdw: "however it has advantages for developers"? And maybe we should also say what the advantages are ;)

README.md
```diff
@@ -33,30 +33,48 @@ We plan to add OAuth 2 support shortly.
...
+From the project root directory, run:
 ```
-cd backend/
-./gradlew bootRun
+./gradlew :monoBuild
 ```
+The `:monoBuild` task is composed of three subtasks, which can be independently invoked if needed:
+- `:frontend:buildClientDev` runs Angular's build command, producing the output files in the default location `frontend/dist`.
+- `:frontend:copyFrontendFiles` copies these files into a directory in `backend` from which Spring will serve static resources.
+- `:backend:bootRun` will compile and run the backend.
```
fdw: Awesome! :+1:

frontend/build.gradle
```diff
@@ -0,0 +1,27 @@
+plugins {
+  id "com.moowork.node" version "1.2.0"
+}
+
+node {
+  version = '10.15.0'
+  npmVersion = '6.4.1'
```
fdw: I think nmp 6.6.0 is the current one. Could you please update? Thank you :)

fdw:
Regarding the tests, it seems that Windows messed up the permissions: https://travis-ci.org/TNG/ApiCenter/jobs/483846622

README.md
fdw: Yeah. Maybe just change the sentence here that usually, you want the single build, but you can do it separately for development?

frontend/package.json
fdw: How about you invest some more time (say half an hour), and if it doesn't work, it's a separate issue?

frontend/build.gradle
catenoid: Sure, but just to check, after running `npm i -g npm@latest` the version is 6.7.0
Is 6.6.0 a LTS version? I can't find much information on npm versions, only [this](https://nodejs.org/en/download/releases/)

.gitignore
catenoid: I believe this is the database created when the app is running without a relational database configured. I deleted it, restarted the app and started adding specifications, and the directory has returned. Can I keep it in the `.gitignore`?

frontend/build.gradle
fdw: Yes, the latest is 6.7.0, released apparantely yesterday ;)

backend/.gitignore
```diff
@@ -85,4 +85,14 @@ gradle-app.setting
 # Apache Lucene
 *.cfe
 *.cfs
-*.si
\ No newline at end of file
+*.si
+
+# Frontend static files
```
fdw: Here, I think src/main/resources/public is enough for now :)

.gitignore
fdw: Yes, I think it makes sense to keep it.

Issue #39, opened 2019-01-22
Add a "null" authentication service


fdw:
As a interested potential user, I'd like to try out ApiCenter without needing Atlassian Crowd so that I can easily see what it is about.

We should offer a pseudo authentication service that allows anyone in without a password, and make it the default.

PR #40, opened 2019-01-24
Add null authentication service

catenoid:
Choose between the services in application.properties, 'crowd' for Atlassian crowd, anything else will route to the null authentication service. When a user logs in the password is ignored.

Integration test `application.properties` also needs this field to avoid a Java IllegalStateException

Issue #39

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Currently ApiCenter can use Atlassian Crowd as an authentication provider, or di
 
 In `backend/src/main/resources/application.properties`, this is configured by the line:
 ```
-auth.service=null
+auth.service=none
 ```
 Writing `auth.service=crowd` configures ApiCenter to use Atlassian crowd. Any other string will remove user authentication.
 
--- a/README.md
+++ b/README.md
@@ -12,8 +12,18 @@ To run ApiCenter, you need `gradle` and angular CLI. The backend itself needs a 
 
 To install angular CLI, enter `npm install -g @angular/cli`
 
-### Atlassian Crowd
-ApiCenter currently requires Atlassian Crowd as an authentication provider. In order to make this work, you need to [add ApiCenter
+### User Authentication
+
+Currently ApiCenter can use Atlassian Crowd as an authentication provider, or disregard user authentication entirely for those who want to try the service without setting up Crowd (in this case, any user can log in with any username, the password field is simply ignored).
+
+In `backend/src/main/resources/application.properties`, this is configured by the line:
+```
+auth.service=null
+```
+Writing `auth.service=crowd` configures ApiCenter to use Atlassian crowd. Any other string will remove user authentication.
+
+#### Atlassian Crowd
+To use Atlassian Crowd as an authentication provider, [add ApiCenter
 as an external application in Atlassian Crowd](https://confluence.atlassian.com/crowd/adding-an-application-18579591.html#AddinganApplication-add) and configure
 the following files within ApiCenter:
 
--- a/README.md
+++ b/README.md
@@ -18,9 +18,9 @@ Currently ApiCenter can use Atlassian Crowd as an authentication provider, or di
 
 In `backend/src/main/resources/application.properties`, this is configured by the line:
 ```
-auth.service=none
+auth.service=crowd
 ```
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -12,8 +12,18 @@ To run ApiCenter, you need `gradle` and angular CLI. The backend itself needs a
...
 
-### Atlassian Crowd
-ApiCenter currently requires Atlassian Crowd as an authentication provider. In order to make this work, you need to [add ApiCenter
+### User Authentication
+
+Currently ApiCenter can use Atlassian Crowd as an authentication provider, or disregard user authentication entirely for those who want to try the service without setting up Crowd (in this case, any user can log in with any username, the password field is simply ignored).
+
+In `backend/src/main/resources/application.properties`, this is configured by the line:
+```
+auth.service=null
```
fdw: I'd rather use "none"

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/Configuration.kt
```diff
@@ -1,17 +1,36 @@
...
 import org.springframework.context.annotation.Configuration
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.beans.factory.annotation.Autowired
+import com.tngtech.apicenter.backend.connector.crowd.service.CrowdAuthenticationService
+import com.tngtech.apicenter.backend.connector.crowd.service.NullAuthenticationService
+import com.tngtech.apicenter.backend.domain.service.ExternalAuthenticationService
 
 @Configuration
-class Configuration {
+class Configuration @Autowired constructor(private val crowdClient: CrowdClient) {
```
fdw: You don't need to do it like this. Methods that are `@Bean` annotated are automatically autowired.

You can write
```
@Bean
fun externalAuthenticationService(crowdClient: CrowdClient) {
...
}
```

But: Now you require even the users without Crowd to have this dependency.

After some googling, I think [ConditionalOnProperty](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/condition/ConditionalOnProperty.html) would be preferable. What do you say?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/crowd/service/CrowdAuthenticationService.kt
```diff
@@ -7,8 +7,7 @@ import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Service
 import java.util.UUID
 
-@Service
-class CrowdAuthenticationService @Autowired constructor(private val crowdClient: CrowdClient) : ExternalAuthenticationService {
+class CrowdAuthenticationService constructor(private val crowdClient: CrowdClient) : ExternalAuthenticationService {
```
fdw: I think you still need the `@Autowired`. Or not?

backend/src/integration-test/resources/application.properties
```diff
@@ -1 +1,2 @@
-jwt.secret=ApiCenterSecuritySecret
\ No newline at end of file
+jwt.secret=ApiCenterSecuritySecret
+auth.service=crowd
```
fdw: Does it make sense to use "crowd" as the default for the tests?

backend/src/main/resources/application.properties
```diff
@@ -1 +1,2 @@
-jwt.secret=### Define a secret that is used to sign JWTs ###
\ No newline at end of file
+jwt.secret=### Define a secret that is used to sign JWTs ###
+auth.service=
```
fdw: How about a comment here listing the options?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/crowd/service/CrowdAuthenticationService.kt
catenoid: Not; if the Configuration class contructor where the CrowdAuthenticationService is created is Autowired. (But that may change.)

fdw: Oh, yes, of course. You're completely right :)

PR #42, opened 2019-01-29
Set request URL for combined and separate builds

catenoid:
HTTP requests made to the backend now use an address sourced from `window.location.origin` instead of a hardcoded `localhost`.

This alone would work for the single build, local and remote, because the requests are sent to the same base URL. However when run seperately (and hence, locally), this is wrong, because Angular's location is on port 4200, and the requests need to go to 8080 (Spring)

### DIFF
--- a/frontend/angular.json
+++ b/frontend/angular.json
@@ -45,6 +45,14 @@
                   "with": "src/environments/environment.prod.ts"
                 }
               ]
+            },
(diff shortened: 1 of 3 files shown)


catenoid:
These changes are incorported into the heroku pull request, since that deals with remote deploys as well as combined / separate.

PR #43, opened 2019-01-29
Add configuration for Heroku deploy

catenoid:
Configuration required to run ApiCenter remotely, on heroku.

After downloading the heroku cli, log in with `heroku login`, and create an app with `heroku create`. Add a postgres database in the heroku web interface under the "Resources" tab (since the in-memory database H2 does not work on heroku).

Deploying to heroku just requires pushing to the master branch of the new git remote, `heroku`. A local branch can be renamed like so: `git push heroku my-branch:master`

After installing PostgreSQL on your computer, a `psql` session can be opened with `heroku pg:psql` and heroku's database can be queried from the command line, if required.


README.md
```diff
@@ -4,51 +4,31 @@
 
 ApiCenter is a repository for all your OpenAPI specifications. You can upload new ones and search them so that finding the one you need is easy.
 
-## Prerequisites
+## How to use
 
-### General
+### Deploy to cloud (heroku)
```
fdw: I'm not sure if I'd put Heroku so prominently in there. After all, ApiCenter is most likely to be deployed internally, and Heroku is not often used for that.

backend/build.gradle
```diff
@@ -49,6 +49,7 @@ dependencies {
     compile "org.springframework.boot:spring-boot-starter-security:$spring_boot_version"
     compile "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
     compile "org.springframework.boot:spring-boot-starter-data-jpa:$spring_boot_version"
+    compile "org.postgresql:postgresql:42.0.0"
```
fdw: So we always use a database now? Does the local execution still work?

backend/src/main/resources/heroku.properties
```diff
@@ -0,0 +1,4 @@
+jwt.secret=${JWT_SECRET}
+spring.datasource.url=${DATABASE_URL}
```
fdw: Is this file used anywhere? Or is it picked up automatically by Heroku?

frontend/angular.json
```diff
@@ -83,6 +95,16 @@
               "src/assets",
               "src/favicon.ico"
             ]
+          },
+          "configurations": {
+            "localhost": {
+              "fileReplacements": [
+                {
+                  "replace": "src/environments/environment.ts",
```
fdw: That is nice :)

backend/build.gradle
catenoid: Yes, it still makes one in memory when running of localhost

backend/src/main/resources/heroku.properties
catenoid: Heroku uses the Procfile to override application.properties with this.

backend/build.gradle
fdw: Okay, interesting :)

PR #48, opened 2019-02-11
API and Angular Routing fixes

### LINKED ISSUE #46: "Not found" specification returns HTTP 500 instead of 404
```
$ curl http://localhost:8080/specifications/65039ab5-b610-4038-9afb-98b45f7fe9fc
{"id":"65039ab5-b610-4038-9afb-98b45f7fe9fc", ... }

$ curl http://localhost:8080/specifications/65039ab5-b610-4038-9afb-98b45f7fe9fd
{"timestamp":1549027851225,"status":500,"error":"Internal Server Error","message":"No value present","path":"/specifications/65039ab5-b610-4038-9afb-98b45f7fe9fd"}
```

### LINKED ISSUE #47: Add a 404 page to the web app
Currently a GET for a non-existent specification in the webapp produces only toolbar + white page, with no indication of what has gone wrong.

blank

catenoid:
Two related changes here to do with the API

## API redefinition
Firstly, modifies Spring config to route all HTTP requests to Angular. Before, every route within Angular's SPA had to be accessed from the root; if the user refreshed the page or sent a static link to someone, Spring would return 404.

Secondly, because both API path requests and Angular routes are now in the same scope / namespace, adds a prefix to all API calls, to distinguish the two services. Updates integration tests. (Also removes '/versions/' from API calls, to make the API vs Angular paths consistent.)

## Add handling for unfound versions / specifications

Before, the findOne methods assumed every request for a version / specification would succeed. When the user gave a non-existent specification ID, the server threw an exception, and 500 was returned to the client.

Now curl requests for non-existent IDs return 404 (fix #46), and in the web frontend, an error banner appears instead of the whole page being blank (fix #47)

### DIFF
--- a/backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
+++ b/backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
@@ -42,6 +42,26 @@ internal class SpecificationControllerIntegrationTest {
     }
 
     @Test
+    fun findAllSpecifications_requiresAuthentication() {
+        mockMvc.perform(get("/api/1.0/specifications"))
+                .andExpect(status().`is`(403))
+    }
+
+    @Test
+    fun findOneSpecification_shouldGetOne() {
+        mockMvc.perform(get("/api/1.0/specifications/b6b06513-d259-4faf-b34b-a216b3daad6a").with(user("user")))
+                .andExpect(status().isOk)
+                .andExpect(content().contentType("application/json;charset=UTF-8"))
+                .andExpect(jsonPath("title", equalTo("Spec1")))
+    }
+
+    @Test
+    fun findOneSpecification_shouldGracefullyFail() {
+        mockMvc.perform(get("/api/1.0/specifications/af0502a2-7410-40e4-90fd-3504f67de1ef").with(user("user")))
+                .andExpect(status().`is`(404))
+    }
+
+    @Test
     fun uploadSpecification_shouldCreateSpecification() {
         mockMvc.perform(
             post("/api/1.0/specifications")
--- a/backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
+++ b/backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
@@ -28,7 +28,7 @@ internal class SpecificationControllerIntegrationTest {
 
     @Test
     fun findAllSpecifications_shouldReturnAllSpecifications() {
-        mockMvc.perform(get("/api/1.0/specifications").with(user("user")))
+        mockMvc.perform(get("/api/v1/specifications").with(user("user")))
             .andExpect(status().isOk)
             .andExpect(content().contentType("application/json;charset=UTF-8"))
             .andExpect(jsonPath("$[0].title", equalTo("Spec1")))
@@ -43,13 +43,13 @@ internal class SpecificationControllerIntegrationTest {
 
     @Test
     fun findAllSpecifications_requiresAuthentication() {
-        mockMvc.perform(get("/api/1.0/specifications"))
-                .andExpect(status().`is`(403))
+        mockMvc.perform(get("/api/v1/specifications"))
+                .andExpect(status().isForbidden)
     }
 
     @Test
     fun findOneSpecification_shouldGetOne() {
-        mockMvc.perform(get("/api/1.0/specifications/b6b06513-d259-4faf-b34b-a216b3daad6a").with(user("user")))
+        mockMvc.perform(get("/api/v1/specifications/b6b06513-d259-4faf-b34b-a216b3daad6a").with(user("user")))
                 .andExpect(status().isOk)
                 .andExpect(content().contentType("application/json;charset=UTF-8"))
                 .andExpect(jsonPath("title", equalTo("Spec1")))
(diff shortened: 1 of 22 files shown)


backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
```diff
@@ -28,7 +28,7 @@ internal class SpecificationControllerIntegrationTest {
 
     @Test
     fun findAllSpecifications_shouldReturnAllSpecifications() {
-        mockMvc.perform(get("/specifications").with(user("user")))
+        mockMvc.perform(get("/api/1.0/specifications").with(user("user")))
```
fdw: Minor version numbers in the API are really unusual. I would rather not use it. Instead something like "v1" so that you can easily see what the number is meant to signify.

backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
```diff
@@ -41,10 +41,30 @@ internal class SpecificationControllerIntegrationTest {
             .andExpect(jsonPath("$[2].versions[1].version", equalTo("1.1")))
     }
 
+    @Test
+    fun findAllSpecifications_requiresAuthentication() {
+        mockMvc.perform(get("/api/1.0/specifications"))
+                .andExpect(status().`is`(403))
```
fdw: There's also `isForbidden()` and so on, which might be easier to read.

backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionControllerIntegrationTest.kt
```diff
@@ -25,15 +25,24 @@ class VersionControllerIntegrationTest {
 
     @Test
     fun findOneVersion_shouldReturnVersion() {
-        mockMvc.perform(get("/specifications/b6b06513-d259-4faf-b34b-a216b3daad6a/versions/v1")
+        mockMvc.perform(get("/api/1.0/specifications/b6b06513-d259-4faf-b34b-a216b3daad6a/v1")
```
fdw: I think I'd rather keep the `/versions/` bit, so that you can get a list of all versions from there.

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/SecurityConfiguration.kt
```diff
@@ -21,22 +21,27 @@ class SecurityConfiguration : WebSecurityConfigurerAdapter() {
...
-            .antMatchers("/open-iconic.*").permitAll()
-            .anyRequest().authenticated()
-            .and()
-            .addFilter(JwtAuthorizationFilter(authenticationManager(), jwtSecuritySecret))
-            .sessionManagement().sessionCreationPolicy(SessionCreationPolicy.STATELESS)
+                .cors()
+                .and()
+                .csrf().disable()
+                .authorizeRequests()
+                .antMatchers("/api/1.0/sessions").permitAll()
```
fdw: There's lots of different matchers. Do we really want the `Ant` ones, or is there something better?

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/WebConfiguration.kt
```diff
@@ -15,6 +16,18 @@ class WebConfiguration {
                 corsRegistry.addMapping("/**").allowedOrigins("*").allowedMethods("*")
                     .allowedHeaders("*")
             }
+
+            override fun addViewControllers(registry: ViewControllerRegistry) {
+                // Redirect to Angular's router
+                // In case you either refresh the page, or want to send a static link to someone
```
fdw: I think the first line of the comment is sufficient

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationController.kt
```diff
@@ -57,8 +57,15 @@ class SpecificationController @Autowired constructor(
         specificationHandler.findAll().map { spec -> specificationDtoMapper.fromDomain(spec) }
 
     @GetMapping("/{specificationId}")
-    fun findSpecification(@PathVariable specificationId: UUID): SpecificationDto =
-        specificationDtoMapper.fromDomain(specificationHandler.findOne(specificationId)!!)
+    @Throws(HttpNotFoundException::class)
+    fun findSpecification(@PathVariable specificationId: UUID): SpecificationDto {
+        val specification = specificationHandler.findOne(specificationId)
+
+        if (specification != null) {
```
fdw: I think you can do something like
```
specification.?let { specificationDtoMapper.fromDomain(it) } ?: throw HttpNotFoundException()
```

backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionControllerIntegrationTest.kt
catenoid: Sorry I don't understand. How does `/versions/` add meaning? The resource path is uniquely identifiable by the specID and version alone. I don't mind changing it, but would like to know why

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationController.kt
catenoid: Awesome 😄 Much neater

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/SecurityConfiguration.kt
catenoid: I was deliberately overcautious here, because (afaiu) the Ant matchers match resources from a number of different roots: `/resources`, `/resources/public`, and the HTTP routes. If you're more liberal with the permitted routes, you can access the properties files, for example.

backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionControllerIntegrationTest.kt
fdw: While I agree that it looks unnecessary, I think I only ever saw REST interfaces with the "parameter-names" added. For example, if I make a GET call to `/specifications/{id}`, I get that resource. If I would make one to `/specifications/{id}/versions` I'd expect to get a list of all the versions. If we don't have the `versions` part anymore, where do I get this list?

catenoid: `/specifications/{specificationId}/versions` was never an API endpoint. (I had not made any edits to api-definition.json until now.) 😕

fdw: Doesn't have to be right now, but at least we have the opportunity.

PR #50, opened 2019-02-21
Support for GraphQL specifications and GraphiQL view

### LINKED ISSUE #17: Add support for GraphQL
[ApiCenter should be open for all kinds of APIs, not just Swagger. One good candidate should be GraphQL](https://graphql.org/): There exists a single specification file for the types. Interestingly, it doesn't appear to contain a name, version or description (this should be added manually when uploading a spec file).
For the detailed view, we could use [Graphiql](https://github.com/graphql/graphiQL) (if the license is okay).

catenoid:
Users can upload GraphQL specifications, and inspect them with GraphiQL.

- Adds an enum field to the Version entity, differentiating OpenAPI and GraphQL, potentially extendible to other spec formats.

Frontend specifically:

- Adds a form UI element which appears when a `.graphql` upload is attemped. Requires the user to add those meta data fields (title, description, version) which are present in the OpenAPI spec, but not in the GraphQL one.

- Adds error handling to file upload component, eg. when a user has not selected a file before pressing "Upload."

- Adds a GraphiQL wrapper Angular component, which renders the GraphiQL React element.

- Adds a parent wrapper component `VersionView` which selectively renders SwaggerUI or GraphiQL, depending on the format, so both have the same HTTP route (ie. `/specifications/{specificationId}/versions/{version}`).

Closes #17 

giql


backend/src/integration-test/resources/data.sql
```diff
@@ -2,8 +2,8 @@ INSERT INTO SPECIFICATION_ENTITY VALUES ('b6b06513d2594fafb34ba216b3daad6a', 're
...
 
-INSERT INTO VERSION_ENTITY VALUES('v1', '{"info": {"title": "Spec1",  "version": "v1", "description": "Description"}}', '2018-01-01 11:00:00.000', 'b6b06513d2594fafb34ba216b3daad6a');
-INSERT INTO VERSION_ENTITY VALUES('v1', '{"info": {"title": "Spec2",  "version": "v1", "description": "Description"}}', '2018-01-01 11:00:00.000', 'f67cb0a6c31b424bbfbbab0e163955ca');
-INSERT INTO VERSION_ENTITY VALUES('v2', '{"info": {"title": "Spec2",  "version": "v2", "description": "Description"}}', '2018-01-01 10:00:00.000', 'f67cb0a6c31b424bbfbbab0e163955ca');
-INSERT INTO VERSION_ENTITY VALUES('1.0', '{"info": {"title": "Spec3",  "version": "1.0", "description": "Description"}}', '2018-01-01 11:00:00.000', 'af0502a2741040e490fd3504f67de1ee');
-INSERT INTO VERSION_ENTITY VALUES('1.1', '{"info": {"title": "Spec3",  "version": "1.1", "description": "Description"}}', '2018-01-01 10:00:00.000', 'af0502a2741040e490fd3504f67de1ee');
\ No newline at end of file
+INSERT INTO VERSION_ENTITY VALUES('v1', '{"info": {"title": "Spec1",  "version": "v1", "description": "Description"}}',   '2018-01-01 11:00:00.000', '0', 'b6b06513d2594fafb34ba216b3daad6a');
```
fdw: Must it be `0`? Can we not use an actual name?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/mapper/configurer/VersionMappingConfigurer.kt
```diff
@@ -17,6 +17,7 @@ class VersionMappingConfigurer constructor(private val versionConverter: Version
         orikaMapperFactory.classMap(VersionEntity::class.java, Version::class.java)
             .field("versionId.version", "version")
             .field("content", "content")
+            .field("language", "language")
```
fdw: Do you need to specifiy that? I thought Orika can map identically called fields automatically.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionController.kt
```diff
@@ -37,7 +37,7 @@ class VersionController constructor(private val versionHandler: VersionHandler,
             logger.info("Specification $specificationId version $version requested as YAML")
             val jsonNodeTree = ObjectMapper().readTree(foundVersion.content)
             val jsonAsYaml = YAMLMapper().writeValueAsString(jsonNodeTree)
-            return versionDtoMapper.fromDomain(Version(foundVersion.version, jsonAsYaml))
+            return versionDtoMapper.fromDomain(Version(foundVersion.version, jsonAsYaml, foundVersion.language))
```
fdw: Does it make sense to render GraphQL as yaml? Is it not something completely different from json and yaml?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/dto/SpecificationFileDto.kt
```diff
@@ -3,8 +3,15 @@ package com.tngtech.apicenter.backend.connector.rest.dto
 import com.fasterxml.jackson.annotation.JsonCreator
 import java.util.UUID
 
+data class SpecificationMetaData constructor(
```
fdw: Are these really things that should be coupled to a `SpecificationFileDto`? Can the title switch between files? If so, would it make sense to have these things as fields in the `SpecificationFileDto`?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/mapper/converter/SpecificationConverter.kt
```diff
@@ -24,19 +25,26 @@ class SpecificationConverter constructor(
...
 
         if (specificationFileDto.fileUrl != null && specificationFileDto.fileUrl != "") {
             fileContent = specificationFileService.retrieveFile(specificationFileDto.fileUrl)
         }
 
-        val content = specificationDataService.parseFileContent(fileContent)
+        val content = if (isGraphQLFile) fileContent else specificationDataService.parseFileContent(fileContent)
         val uuid = specificationFileDto.id ?: UUID.randomUUID()
 
+        val title = metaData?.title?: specificationDataService.readTitle(content)
```
fdw: How about we save this data for OpenApi too? Then we wouldn't have to do this here.

frontend/src/app/header/header.component.html
```diff
@@ -2,7 +2,7 @@
   <span class="navbar-brand">ApiCenter</span>
 
   <ul class="navbar-nav mr-auto" *ngIf="username">
-    <li class="nav-item"><a routerLink="/">Home</a></li>
+    <li class="nav-item"><a href="/">Home</a></li>
```
fdw: Why this change?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/mapper/converter/SpecificationConverter.kt
```diff
@@ -24,19 +25,26 @@ class SpecificationConverter constructor(
         mappingContext: MappingContext?
     ): Specification {
         var fileContent = specificationFileDto.fileContent ?: ""
+        val metaData = specificationFileDto.metaData
+        val isGraphQLFile = metaData != null
```
fdw: I think I'd rather note this explicitely somewhere.

frontend/src/app/specification-form/specification-form.component.html
```diff
@@ -6,27 +6,43 @@
   </div>
   <div class="row top-buffer">
     <div class="col">
-      <input type="radio" name="type-radio" (click)="changeTypeRadio()" [checked]="!manualDisabled"/>
-      Upload file manually ...
+      <input type="radio" name="type-radio" (click)="changeTypeRadio()" [checked]="!remoteUploadSelected"/>
+      Upload a file from your computer
```
fdw: I like this change :)

frontend/src/app/specification-form/specification-form.component.html
```diff
@@ -6,27 +6,43 @@
...
     <div class="col">
       <p>
-        <input type="file" placeholder="Upload Swagger Specification" accept=".yml,.json,.yaml" ngModel
-               (change)="getFile($event)" [attr.disabled]="manualDisabled ? '' : null">
+        <input type="file" placeholder="Upload Swagger Specification" accept=".yml,.json,.yaml,.graphql" ngModel
+               (change)="onLocalFileChange($event)" [attr.disabled]="remoteUploadSelected ? '' : null">
       </p>
     </div>
+    <div class="col" *ngIf="isGraphQLFile">
+      Additionally required fields for GraphQL upload:
```
fdw: I wouldn't write here that it is necessary only for GraphQL. Rather something like "Please add a description of the service:"

backend/src/integration-test/resources/data.sql
catenoid: Unfortunatly not, because that is what I first tried. I think this is just how Kotlin's enums are represented. Would be glad to be proven wrong though, as it's not clear what the number represents.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionController.kt
catenoid: GraphQL isn't YAML, and this `if` block is not reached for GraphQL. But the `Version` constructor now requires a language parameter. Maybe I haven't understood your question?

frontend/src/app/specification-form/specification-form.component.html
```diff
@@ -6,27 +6,43 @@
...
+      <div *ngFor="let fieldName of objectKeys(metaData)">
+        <div class="row">
+          <input type="text" [placeholder]="fieldName.charAt(0).toUpperCase() + fieldName.slice(1)" [(ngModel)]="metaData[fieldName]"/>
+        </div>
+      </div>
+      <button (click)="addServerInputField()">Add</button>
+      <button (click)="removeServerInputField()">Remove</button>
+      <p>API endpoints:</p>
```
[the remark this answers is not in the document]
catenoid: Yes it can, and they also have descriptions https://swagger.io/docs/specification/api-host-and-base-path/

frontend/src/app/specification-version/graphiql-wrapper.component.ts
```diff
@@ -0,0 +1,70 @@
...
+    invariant(node, `Node '${this.graphiql} not found!`);
+    return node;
+  }
+
+  private isMounted(): boolean {
+    return !!this.graphiql && !!this.specification;
+  }
+
+  public graphQLFetcher(graphQLParams) {
+    return fetch('http://localhost:4000/graphql', {
```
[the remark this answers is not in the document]
catenoid: No 😄 The server urls are sort of 'under construction' atm, because I thought it would be better if I could get your feedback on the rest earlier.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/mapper/configurer/VersionMappingConfigurer.kt
catenoid: Yes, they are required. (When omited all file uploads fail, just checked.)

frontend/src/app/header/header.component.html
catenoid: Fixed now 🎉 So the change is removed. (GraphiQL messed up the router history, so neither this Home link nor the browser back button worked. Since figured out why. This was a workaround, so the link still worked.)

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/VersionEntity.kt
```diff
@@ -1,32 +1,31 @@
...
-    @Field @Column(columnDefinition = "TEXT") val content: String,
-    @Field @Column(columnDefinition = "TEXT") val language: APILanguage,
-    @ContainedIn @ManyToOne @MapsId("specificationId") var specification: SpecificationEntity?,
-    var created: Date?
+        @EmbeddedId @IndexedEmbedded val versionId: VersionId,
+        @Field @Column(columnDefinition = "TEXT") val content: String,
+        @Field @Column(columnDefinition = "TEXT") val title: String,
+        @Field @Column(columnDefinition = "TEXT") val description: String,
+        @Field @Column(columnDefinition = "TEXT") val language: ApiLanguage,
+        @Field @Column(columnDefinition = "TEXT") val servers: String,
```
fdw: This is a list of servers, right? Please do not store a list as a single item in the DB.
For now, we can probably only have one server.

Later, we need a set of them, but correlated with environments and stuff.

frontend/src/app/specification-version/graphiql-wrapper.component.ts
```diff
@@ -40,21 +40,24 @@ export class GraphiQLWrapperComponent extends VersionViewComponent implements On
...
-    return fetch('http://localhost:4000/graphql', {
-      method: 'POST',
-      headers: { 'Content-Type': 'application/json' },
-      body: JSON.stringify(graphQLParams)
-    }).then(response => response.json())
-      .catch(() => 'An error occurred with the network request. ' +
-        '(Check the GraphQL service is running, and accepts POST requests)');
+  public graphQLFetcher(apiEndpoint?: string) {
+    return (graphQLParams) => {
+      return fetch(apiEndpoint ? apiEndpoint : 'http://localhost:4000/graphql', { // As a fallback, use express-graphql's default
```
fdw: How about we use a new endpoint in the backend as a fallback that will only return an error?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/VersionEntity.kt
```diff
@@ -21,9 +21,9 @@ data class VersionEntity(
         @EmbeddedId @IndexedEmbedded val versionId: VersionId,
         @Field @Column(columnDefinition = "TEXT") val content: String,
         @Field @Column(columnDefinition = "TEXT") val title: String,
-        @Field @Column(columnDefinition = "TEXT") val description: String,
+        @Field @Column(columnDefinition = "TEXT") val description: String?,
         @Field @Column(columnDefinition = "TEXT") val language: ApiLanguage,
-        @Field @Column(columnDefinition = "TEXT") val servers: String,
+        @Field @Column(columnDefinition = "TEXT") val server: String?,
```
fdw: One last idea: How about calling this "endpoint" instead?

catenoid: endpointUrl, to make clear what the contents represent?

frontend/src/app/specification-version/graphiql-wrapper.component.ts
catenoid: I much prefer this idea 😎

PR #53, opened 2019-02-27
Add HTTP response codes with @ControllerAdvice

catenoid:
What were bad client requests (eg. no title given in an OpenAPI specification) were all eliciting 500 code responses from the server. This also printed long stack traces in the kotlin logger.

This uses Spring's `@ControllerAdvice` to return 4xx client error codes, with the added advantage of not printing stack traces. Some thrown exceptions are made more specific. (Also, this overrides the standard Spring error response object to have a more meaningful ISO timestamp, the default one is just an integer with no indication of how to interpret its value)

The error messages are all still returned and displayed to the client correctly in the frontend.

### DIFF
--- a/backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
+++ b/backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
@@ -6,12 +6,25 @@ import org.springframework.web.bind.annotation.ControllerAdvice
 import org.springframework.web.bind.annotation.ExceptionHandler
 import java.time.Instant
 import java.time.format.DateTimeFormatter
+import mu.KotlinLogging
+import java.util.*
+
+private val logger = KotlinLogging.logger {  }
+class SpecificationNotFoundException(val specificationId: UUID, val version: String? = "") : RuntimeException()
+class SpecificationParseException(val msg: String) : RuntimeException()
+class VersionAlreadyExistsException(val specificationTitle: String) : RuntimeException()
+
+data class ErrorMessage(
+        val message: String,
+        val timestamp: String = DateTimeFormatter.ISO_INSTANT.format(Instant.now())
+)
 
 @ControllerAdvice
-class RestResponseExceptionHandler() {
+class RestResponseExceptionHandler {
 
     @ExceptionHandler(SpecificationNotFoundException::class)
-    fun handleNotFound(): ResponseEntity<ErrorMessage> {
+    fun handleNotFound(exc: SpecificationNotFoundException): ResponseEntity<ErrorMessage> {
+        logger.info("Specification ${exc.specificationId} ${exc.version} not found")
         return ResponseEntity(
                 ErrorMessage("Specification not found"),
                 null,
@@ -19,8 +32,8 @@ class RestResponseExceptionHandler() {
         )
     }
 
-    @ExceptionHandler(SpecificationParseFailureException::class)
-    fun handleBadRequest(exc: SpecificationParseFailureException): ResponseEntity<ErrorMessage> {
+    @ExceptionHandler(SpecificationParseException::class)
+    fun handleBadRequest(exc: SpecificationParseException): ResponseEntity<ErrorMessage> {
         return ResponseEntity(
                 ErrorMessage(exc.msg),
                 null,
@@ -33,16 +46,8 @@ class RestResponseExceptionHandler() {
         return ResponseEntity(
                 ErrorMessage("A specification with the same version already exists for ${exc.specificationTitle}"),
                 null,
-                HttpStatus.UNPROCESSABLE_ENTITY
+                HttpStatus.CONFLICT
         )
     }
 }
 
-class SpecificationNotFoundException : RuntimeException()
-class SpecificationParseFailureException(val msg: String) : RuntimeException()
-class VersionAlreadyExistsException(val specificationTitle: String) : RuntimeException()
-
-data class ErrorMessage(
-        val message: String,
-        val timestamp: String = DateTimeFormatter.ISO_INSTANT.format(Instant.now())
-)
--- a/backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
+++ b/backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
@@ -19,7 +19,7 @@ data class ErrorMessage(
         val timestamp: String = DateTimeFormatter.ISO_INSTANT.format(Instant.now())
 )
 
-fun responseBuilder(userMessage: String, status: HttpStatus): ResponseEntity<ErrorMessage> {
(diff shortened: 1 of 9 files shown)


backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.web.bind.annotation.ExceptionHandler
+import java.time.Instant
+import java.time.format.DateTimeFormatter
+
+@ControllerAdvice
+class RestResponseExceptionHandler() {
+
+    @ExceptionHandler(SpecificationNotFoundException::class)
+    fun handleNotFound(): ResponseEntity<ErrorMessage> {
+        return ResponseEntity(
```
fdw: What do you think about returning an actual object that could be serialized to JSON, instead of only the message as a string? Often, you can then add both a machine-readable error code (f.e. `SPECIFICATION_NOT_FOUND`) and a human-readable message with more details ("could not find the specification $name").
Additionally, I've seen the concept of logrefs: A UUID that is passed to the user and noted in the log so that debugging the problem gets easier.

fdw: Also, I think we should log the exception.

backend/src/test/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataServiceTest.kt
```diff
@@ -53,15 +54,15 @@ class SpecificationDataServiceTest {
     @Test
     fun readTitle_shouldFailWhenNoTitleIsGiven() {
         assertThatThrownBy { specificationDataService.readTitle(SWAGGER_SPECIFICATION_WITHOUT_TITLE) }.isInstanceOf(
-            IllegalArgumentException::class.java
-        ).hasMessage("No title given in content.")
+            SpecificationParseFailureException::class.java
```
fdw: I'm not sure if "failure" and "exception" are both needed ;)

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
catenoid: This is already the case, it's just the second field is set automatically, ie. the client receives `{"message":"Specification needs a title","timestamp":"2019-02-27T14:25:05.508Z"}`. I just put the bare minumum fields in this commit.

fdw: What do you think about the machine-readable error code and a separate message?

catenoid: I've nothing against it, but I'm more motivated by only what fields the application needs in it's current state. Is there a convention / spec for API responses you could point me to, if these fields are expected?

fdw: No, there is no convetion. However, I've learned the need for both fields in previous projects, as a machine readable code cannot contain details, but you need them for actual debugging. So, for me at least, it's a good practice that I'd like to spread.

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
```diff
@@ -6,21 +6,34 @@ import org.springframework.web.bind.annotation.ControllerAdvice
...
+)
 
 @ControllerAdvice
-class RestResponseExceptionHandler() {
+class RestResponseExceptionHandler {
 
     @ExceptionHandler(SpecificationNotFoundException::class)
-    fun handleNotFound(): ResponseEntity<ErrorMessage> {
+    fun handleNotFound(exc: SpecificationNotFoundException): ResponseEntity<ErrorMessage> {
+        logger.info("Specification ${exc.specificationId} ${exc.version} not found")
```
fdw: Please also log the actual exception, as the stack trace is in there.

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
```diff
@@ -6,21 +6,34 @@ import org.springframework.web.bind.annotation.ControllerAdvice
 import org.springframework.web.bind.annotation.ExceptionHandler
 import java.time.Instant
 import java.time.format.DateTimeFormatter
+import mu.KotlinLogging
+import java.util.*
+
+private val logger = KotlinLogging.logger {  }
+class SpecificationNotFoundException(val specificationId: UUID, val version: String? = "") : RuntimeException()
```
fdw: These exceptions must be defined somewhere in the domain layer, as they are (obviously) domain exceptions.

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
```diff
@@ -1,53 +1,48 @@
...
-class VersionAlreadyExistsException(val specificationTitle: String) : RuntimeException()
 
 data class ErrorMessage(
-        val message: String,
+        val httpReasonPhrase: String,
+        val userMessage: String,
         val timestamp: String = DateTimeFormatter.ISO_INSTANT.format(Instant.now())
 )
 
+fun responseBuilder(userMessage: String, status: HttpStatus): ResponseEntity<ErrorMessage> {
```
fdw: I'd rather not call this a "builder", as it's more a factory method.

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
```diff
@@ -1,53 +1,48 @@
...
+        null,
+        status
+    )
+}
+
 @ControllerAdvice
 class RestResponseExceptionHandler {
 
     @ExceptionHandler(SpecificationNotFoundException::class)
     fun handleNotFound(exc: SpecificationNotFoundException): ResponseEntity<ErrorMessage> {
```
fdw: You know that I'm a fan of long, explicit names like "exception", right? ;)

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
```diff
@@ -1,53 +1,48 @@
...
-        logger.info("Specification ${exc.specificationId} ${exc.version} not found")
-        return ResponseEntity(
-                ErrorMessage("Specification not found"),
-                null,
-                HttpStatus.NOT_FOUND
-        )
+        val detailedMessage = "Specification ${exc.specificationId} ${exc.version} not found"
+        logger.info(detailedMessage, exc)
```
fdw: If you inline the string variable, it should be a bit more performant. Also, I think an `error` is more appriopriate :)

PR #60, opened 2019-03-22
Group specifications by ID instead of title, and x-api-id support

### LINKED ISSUE #16: API Version Grouping Based on API Identifier
We are using UUIDs as API identifiers as suggested by the Zalando API guidelines (https://opensource.zalando.com/restful-api-guidelines/#215). We would like to group APIs based on this identifier (currently no identifier is supported and every submitted API spec leads to a new API created, even on version updates). We will create a pull request if you also support this idea.

catenoid:
Groups specifications by ID instead of by title.

Either the ID is a field in the specification file DTO when the specification is uploaded, or in the specification body under `info/x-api-id`. Non-OpenAPI formatted specifications (eg. GraphQL) will only be able to use the former. If there is a mismatch between the ID in the upload content and the ID in the URL used for the upload request, the upload is rejected.

To allow for arbitrary user defined IDs, changes the ID type within ApiCenter from a UUID, to a class which wraps a String. Because JPA only allows primitive Java types as @Id fields, this is still a String, with the wrapping / unwrapping occurring when the data is stored / retrieved.

IDs are restricted to alphanumeric strings with hyphens and underscores, validation occurs in the wrapper class constructor. This may be unnecessarily restrictive, but without any check, a specification with eg. `x-api-id: unique/identifier` can be uploaded, but never viewed or deleted. Clicking on the links within the web interface will involve a different string literal (`http://localhost:4200/specifications/unique%2Fidentifier/0.1.8`)

Some of the columns of type String in the integration test database population file were in the wrong order, however this has only just now become apparent with the addition of the tests in this PR.

Also runs IntelliJ's "Optimise Imports" (to get rid of the `java.util.UUID` references), hence those diffs.

Closes #16

### DIFF
--- a/backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
+++ b/backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
@@ -10,13 +10,8 @@ import org.springframework.security.test.web.servlet.request.SecurityMockMvcRequ
 import org.springframework.security.test.web.servlet.request.SecurityMockMvcRequestPostProcessors.user
 import org.springframework.test.context.junit4.SpringRunner
 import org.springframework.test.web.servlet.MockMvc
-import org.springframework.test.web.servlet.request.MockMvcRequestBuilders.delete
-import org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get
-import org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post
-import org.springframework.test.web.servlet.request.MockMvcRequestBuilders.put
-import org.springframework.test.web.servlet.result.MockMvcResultMatchers.content
-import org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath
-import org.springframework.test.web.servlet.result.MockMvcResultMatchers.status
+import org.springframework.test.web.servlet.request.MockMvcRequestBuilders.*
+import org.springframework.test.web.servlet.result.MockMvcResultMatchers.*
 
 @RunWith(SpringRunner::class)
 @SpringBootTest
@@ -111,7 +106,7 @@ internal class SpecificationControllerIntegrationTest {
                 .content(
                     """
                         | {
-                        |   "fileContent": "{\"info\": {\"title\": \"Spec1\",\"version\": \"v2\"}}"
+                        |   "fileContent": "{\"info\": {\"title\": \"Spec1\",\"version\": \"v2\",\"x-api-id\": \"unique-identifier\"}}"
                         | }
                     """.trimMargin()
                 )
@@ -121,13 +116,15 @@ internal class SpecificationControllerIntegrationTest {
             .andExpect(jsonPath("$.versions[0].metadata.version", equalTo("v2")))
 
         mockMvc.perform(
-            get("/api/v1/specifications/b6b06513-d259-4faf-b34b-a216b3daad6a")
+            get("/api/v1/specifications/unique-identifier")
                 .with(user("user"))
                 .with(csrf())
         )
             .andExpect(jsonPath("$.title", equalTo("Spec1")))
             .andExpect(jsonPath("$.versions[0].metadata.version", equalTo("v2")))
+            .andExpect(jsonPath("$.versions[0].metadata.title", equalTo("Spec1")))
             .andExpect(jsonPath("$.versions[1].metadata.version", equalTo("v1")))
+            .andExpect(jsonPath("$.versions[1].metadata.title", equalTo("Spec4")))
     }
 
     @Test
@@ -155,6 +152,43 @@ internal class SpecificationControllerIntegrationTest {
         )
             .andExpect(jsonPath("$.title", equalTo("NewSpec")))
             .andExpect(jsonPath("$.versions[0].metadata.version", equalTo("vX")))
+    }
+
+    @Test
+    fun updateSpecification_shouldRejectIdMismatchInXApiId() {
+        mockMvc.perform(
+                put("/api/v1/specifications/b6b06513-d259-4faf-b34b-a216b3daad6a")
+                        .with(user("user"))
+                        .with(csrf())
+                        .contentType("application/json")
+                        .content(
+                                """
+                            | {
+                            |   "fileContent": "{\"info\": {\"title\": \"NewSpec\",\"version\": \"vX\", \"x-api-id\": \"does-not-match\"}}"
+                            | }
+                            """.trimMargin()
+                        )
+        )
+                .andExpect(status().isBadRequest)
+    }
+
+    @Test
+    fun updateSpecification_shouldRejectIdMismatchInDto() {
+        mockMvc.perform(
+                put("/api/v1/specifications/b6b06513-d259-4faf-b34b-a216b3daad6a")
+                        .with(user("user"))
+                        .with(csrf())
+                        .contentType("application/json")
+                        .content(
+                                """
+                            | {
+                            |   "id": "mismatch",
+                            |   "fileContent": "{\"info\": {\"title\": \"NewSpec\",\"version\": \"vX\"}}"
+                            | }
+                            """.trimMargin()
+                        )
+        )
+                .andExpect(status().isBadRequest)
     }
 
     @Test
--- a/backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
+++ b/backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
@@ -48,7 +48,7 @@ class RestResponseExceptionHandler {
                     "A-Z characters, underscores and hyphens", HttpStatus.BAD_REQUEST)
 
     @ExceptionHandler(SpecificationUploadUrlMismatch::class)
-    fun handleUnacceptableId(exception: SpecificationUploadUrlMismatch) =
+    fun handleUrlMismatch(exception: SpecificationUploadUrlMismatch) =
             responseFactory("The API ID used in the upload URL (${exception.urlPathId}) " +
                     "is not the same as the API ID used in the specification body (${exception.userDefinedId})", HttpStatus.BAD_REQUEST)
 }
--- a/backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
+++ b/backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
@@ -17,7 +17,7 @@ data class ErrorMessage(
         val timestamp: String = DateTimeFormatter.ISO_INSTANT.format(Instant.now())
 )
 
-fun responseFactory(userMessage: String, status: HttpStatus): ResponseEntity<ErrorMessage> {
+fun makeResponseEntity(userMessage: String, status: HttpStatus): ResponseEntity<ErrorMessage> {
     return ResponseEntity(
         ErrorMessage(status.reasonPhrase, userMessage),
         null,
@@ -31,25 +31,25 @@ class RestResponseExceptionHandler {
     @ExceptionHandler(SpecificationNotFoundException::class)
     fun handleNotFound(exception: SpecificationNotFoundException): ResponseEntity<ErrorMessage> {
         logger.info("Specification ${exception.specificationId} ${exception.version} not found", exception)
-        return responseFactory("Specification not found", HttpStatus.NOT_FOUND)
+        return makeResponseEntity("Specification not found", HttpStatus.NOT_FOUND)
     }
 
     @ExceptionHandler(SpecificationParseException::class)
     fun handleBadRequest(exception: SpecificationParseException) =
-            responseFactory(exception.userMessage, HttpStatus.BAD_REQUEST)
+            makeResponseEntity(exception.userMessage, HttpStatus.BAD_REQUEST)
 
     @ExceptionHandler(VersionAlreadyExistsException::class)
     fun handleVersionAlreadyExists(exception: VersionAlreadyExistsException) =
-            responseFactory("A specification with the same version already exists for ${exception.specificationTitle}", HttpStatus.CONFLICT)
+            makeResponseEntity("A specification with the same version already exists for ${exception.specificationTitle}", HttpStatus.CONFLICT)
 
     @ExceptionHandler(InvalidSpecificationIdException::class)
     fun handleUnacceptableId(exception: InvalidSpecificationIdException) =
-            responseFactory("The API ID supplied (${exception.userDefinedId}) should only contain numbers, " +
+            makeResponseEntity("The API ID supplied (${exception.userDefinedId}) should only contain numbers, " +
                     "A-Z characters, underscores and hyphens", HttpStatus.BAD_REQUEST)
 
     @ExceptionHandler(MismatchedSpecificationIdException::class)
     fun handleServiceIdMismatch(exception: MismatchedSpecificationIdException) =
-            responseFactory("The API ID used in the upload URL (${exception.urlPathId}) " +
+            makeResponseEntity("The API ID used in the upload URL (${exception.urlPathId}) " +
                     "is not the same as the API ID used in the specification body (${exception.userDefinedId})", HttpStatus.BAD_REQUEST)
 }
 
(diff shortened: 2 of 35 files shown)


backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/repository/SpecificationRepository.kt
```diff
@@ -3,11 +3,11 @@ package com.tngtech.apicenter.backend.connector.database.repository
...
+import java.util.*
 
 @Repository
-interface SpecificationRepository : CrudRepository<SpecificationEntity, UUID> {
+interface SpecificationRepository : CrudRepository<SpecificationEntity, String> {
 
-    fun existsByTitle(title: String): Boolean
-    fun findByTitle(title: String): SpecificationEntity
+    override fun existsById(id: String): Boolean
+    override fun findById(id: String): Optional<SpecificationEntity>
```
fdw: Don't we also need "findByTitle" for the search?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
```diff
@@ -24,26 +24,29 @@ class SpecificationDatabaseService constructor(
     override fun save(specification: Specification) {
         val specificationEntity = specificationEntityMapper.fromDomain(specification)
 
-        if (specificationRepository.existsByTitle(specificationEntity.title)) {
-            val existingEntity = specificationRepository.findByTitle(specificationEntity.title)
-
+        val entityToStore = specificationRepository.findById(specificationEntity.id).map {
+            existingEntity ->
```
fdw: I find this a bit long for a lambda. Either extract it into a nice method or restructure it to not require a lambda (or at least such a long one).

By the way, instead of naming the parameter, you can also just use `it`.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
```diff
@@ -24,26 +24,29 @@ class SpecificationDatabaseService constructor(
...
             val versions = existingEntity.versions + specificationEntity.versions
-            val newEntity = SpecificationEntity(existingEntity.id, existingEntity.title, existingEntity.description, versions, existingEntity.remoteAddress)
+            SpecificationEntity(
```
fdw: Why don't we have a "appendVersion" method on `SpecificationEntity`?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationController.kt
```diff
@@ -37,15 +29,22 @@ class SpecificationController @Autowired constructor(
         return specificationDtoMapper.fromDomain(specification)
     }
 
-    @PutMapping("/{specificationId}")
-    fun updateSpecification(@RequestBody specificationFileDto: SpecificationFileDto, @PathVariable specificationId: String): SpecificationDto {
+    @PutMapping("/{urlPathId}")
+    fun updateSpecification(@RequestBody specificationFileDto: SpecificationFileDto, @PathVariable urlPathId: String): SpecificationDto {
+
+        val specificationId = specificationFileDto.id?.let {
```
fdw: I'd extract this into a small, private method.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationController.kt
```diff
@@ -37,15 +29,22 @@ class SpecificationController @Autowired constructor(
         return specificationDtoMapper.fromDomain(specification)
     }
 
-    @PutMapping("/{specificationId}")
-    fun updateSpecification(@RequestBody specificationFileDto: SpecificationFileDto, @PathVariable specificationId: String): SpecificationDto {
+    @PutMapping("/{urlPathId}")
```
fdw: That variable name makes it sound like it's the ID for a urlPath. How about "specificationIdFromPath"?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationController.kt
```diff
@@ -59,19 +58,19 @@ class SpecificationController @Autowired constructor(
         specificationHandler.findAll().map { spec -> specificationDtoMapper.fromDomain(spec) }
 
     @GetMapping("/{specificationId}")
-    fun findSpecification(@PathVariable specificationId: UUID): SpecificationDto {
+    fun findSpecification(@PathVariable specificationId: String): SpecificationDto {
```
fdw: A completetly different idea: "Stringly typed" is a bad pattern, because you don't have the type safety anymore - anything can be a string. Do you think a small value object "SpecificationId" that just wraps the string makes sense?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/VersionController.kt
```diff
@@ -25,27 +23,28 @@ class VersionController constructor(private val versionHandler: VersionHandler,
...
-            throw SpecificationNotFoundException(specificationId, version)
-        } else if (accept == MEDIA_TYPE_YAML) {
+        val foundVersion = versionHandler.findOne(specificationId, version) ?: throw SpecificationNotFoundException(specificationId, version)
+
+        val convertedVersion = if (accept == MEDIA_TYPE_YAML) {
             logger.info("Specification $specificationId version $version requested as YAML")
             val jsonNodeTree = ObjectMapper().readTree(foundVersion.content)
             val jsonAsYaml = YAMLMapper().writeValueAsString(jsonNodeTree)
-            return versionDtoMapper.fromDomain(Version(jsonAsYaml, foundVersion.metadata))
+            Version(jsonAsYaml, foundVersion.metadata)
```
fdw: I'm not a fan of the short-hand `val blah = if () { [a lot of stuff happening] suddenly an object appears }`. Again, a good alternative is to extract another small method.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/mapper/converter/SpecificationConverter.kt
```diff
@@ -34,7 +36,19 @@ class SpecificationConverter constructor(
         }
 
         val content = if (isGraphQLFile) fileContent else specificationDataService.parseFileContent(fileContent)
-        val uuid = specificationFileDto.id ?: UUID.randomUUID()
+
+        val userDefinedId = specificationDataService.readId(content)
+        val urlPathId = specificationFileDto.id
+
+        if (userDefinedId?.let { userId -> !isAcceptable(userId) } ?: run { false } ) {
```
fdw: Is this better readable than `if(userDefinedId != null && !isAcceptable(userDefinedId))`?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/mapper/converter/SpecificationConverter.kt
```diff
@@ -53,4 +67,14 @@ class SpecificationConverter constructor(
             specificationFileDto.fileUrl
         )
     }
+
+    private fun isAcceptable(userId: String): Boolean {
```
fdw: "containsInvalidCharacters" could save us the comment ;)

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataService.kt
```diff
@@ -72,4 +71,11 @@ class SpecificationDataService @Autowired constructor(
         } catch (exception: PathNotFoundException) {
             null
         }
+
+    fun readId(json: String): String? =
```
fdw: "extractId"?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/repository/SpecificationRepository.kt
catenoid: No, these weren't for searching; the query applies to multiple fields, like so
```
 specificationQueryBuilder.keyword().onFields("title", "description", "versions.content").matching(searchString)
                            .createQuery()
```

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
```diff
@@ -43,5 +41,15 @@ class RestResponseExceptionHandler {
     @ExceptionHandler(VersionAlreadyExistsException::class)
     fun handleVersionAlreadyExists(exception: VersionAlreadyExistsException) =
             responseFactory("A specification with the same version already exists for ${exception.specificationTitle}", HttpStatus.CONFLICT)
+
+    @ExceptionHandler(UnacceptableUserDefinedApiId::class)
+    fun handleUnacceptableId(exception: UnacceptableUserDefinedApiId) =
```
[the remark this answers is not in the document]
catenoid: Yes, now I think it over, "Acceptable" in an HTTP context usually refers to the `Accept` header negotiation?

backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
```diff
@@ -157,6 +154,43 @@ internal class SpecificationControllerIntegrationTest {
             .andExpect(jsonPath("$.versions[0].metadata.version", equalTo("vX")))
     }
 
+    @Test
+    fun updateSpecification_shouldRejectIdMismatchInXApiId() {
```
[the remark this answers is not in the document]
catenoid: Not sure. On one hand, I like how it is more legible, on the other hand, imo the expected behaviour is so closely coupled to / dependent on the scenario.So it's not much of a 'classification hierarchy' in the way the underscore partitions the tests by the method they apply to.

backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
```diff
@@ -121,13 +116,15 @@ internal class SpecificationControllerIntegrationTest {
...
 
         mockMvc.perform(
-            get("/api/v1/specifications/b6b06513-d259-4faf-b34b-a216b3daad6a")
+            get("/api/v1/specifications/unique-identifier")
                 .with(user("user"))
                 .with(csrf())
         )
             .andExpect(jsonPath("$.title", equalTo("Spec1")))
             .andExpect(jsonPath("$.versions[0].metadata.version", equalTo("v2")))
+            .andExpect(jsonPath("$.versions[0].metadata.title", equalTo("Spec1")))
```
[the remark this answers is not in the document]
catenoid: Yes, if I understand you right, this is the order the specifications would be displayed in the `specification-overview`, by upload date. Which reminds me of a related question: do we ever want to override this, and allow a service maintainer to sort by the version string? I thought it might be useful. For example if someone is uploading a large number of existing specifications at once, and doesn't want to take the time to manually sort.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataService.kt
catenoid: "getId"? imo most have an idea of what a "getter" does, whereas it wouldn't be immediately clear to me what an "extractor" does on top of that. I'm not a fan of "read", to me something called a "reader" usually involves file I/O?

fdw: I'm not a fan of "get", because it covers such a wide area. For me, it should only be used in the narrow cases of returning the value if a class field (or similar things). Here, the json parameter is parsed and only a part of it is returned; for me, that is exactly the idea behing "extract".

backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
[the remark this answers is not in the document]
fdw: I agree, but now the expectation comes before the scenario, which is why I suggested the change. It's surely not ideal, but I think it's a bit better. But I'm open to other ideas :)

backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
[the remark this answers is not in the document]
fdw: Yes and no: It should always be sorted by the version, and not be able to be overridden.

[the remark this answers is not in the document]
catenoid: Alphabetically sorted? It currently isn't: `@OrderBy("created DESC") val versions: List<VersionEntity>`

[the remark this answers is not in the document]
fdw: Hm. Ideally, by version numbers, so that 4.1.0.RELEASE is better than 4.1.0.RC2 and also 4.1.0.BETA. But let's keep this out of scope for now. I'd suggest alphabetically descending, that should handle all the numbers well enough.

Another thing: Do we actually send all the versions on the first load? Because I think we should only the send the most current one and let it reload the other ones later, if the user actually wants them.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
```diff
@@ -22,29 +22,24 @@ class SpecificationDatabaseService constructor(
...
 
     override fun findAll(): List<Specification> = specificationRepository.findAll().map { spec -> specificationEntityMapper.toDomain(spec) }
 
-    override fun findOne(id: UUID): Specification? =
-        specificationRepository.findById(id).orElse(null)?.let { spec -> specificationEntityMapper.toDomain(spec) }
+    override fun findOne(id: ServiceId): Specification? =
+        specificationRepository.findById(id.id).orElse(null)?.let { spec -> specificationEntityMapper.toDomain(spec) }
 
-    override fun delete(id: UUID) = specificationRepository.deleteById(id)
+    override fun delete(id: ServiceId) = specificationRepository.deleteById(id.id)
```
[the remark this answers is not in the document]
catenoid: Unfortunately not, that was my first attempt. It's because `@Id` annotated fields must be primitive types. However I just found out [here](https://www.baeldung.com/spring-mvc-custom-data-binder) that (in response to a different comment of yours) I can automatically convert Spring's `@PathVariables` to `ServiceId`s

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/SpecificationEntity.kt
```diff
@@ -3,23 +3,29 @@ package com.tngtech.apicenter.backend.connector.database.entity
...
     @IndexedEmbedded @OneToMany(
         mappedBy = "specification",
         cascade = [CascadeType.ALL]
     ) @OrderBy("created DESC") val versions: List<VersionEntity>,
     val remoteAddress: String?
-)
\ No newline at end of file
+) {
+
+    fun pureUpdate(other: SpecificationEntity): SpecificationEntity {
```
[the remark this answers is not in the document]
catenoid: 1. Correct, in my mind the SpecificationEntity is only used like a transfer object passed to the repository, it doesn't really "store" data in a persistence sense. So it seemed misleading to have a method modify the object directly, which is why I made it "pure" / "immutable" rather than "in-place". </br></br>2. `other` is intended as the "newer" spec, the one which overrides the title / description with more recent content. Because the `id` is not supposed to be changed, it is preserved </br></br>3. Not sure. I assume the Version ordering is determined by queries to the repository in the `@OrderBy` annotation? I don't know if the ordering of the `VersionEntity`s here takes effect.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/mapper/configurer/VersionMappingConfigurer.kt
```diff
@@ -16,11 +16,11 @@ class VersionMappingConfigurer constructor(private val versionConverter: Version
 
         orikaMapperFactory.classMap(VersionEntity::class.java, Version::class.java)
             .field("versionId.version", "metadata.version")
-            .field("content", "content")
             .field("title", "metadata.title")
             .field("description", "metadata.description")
             .field("language", "metadata.language")
             .field("endpointUrl", "metadata.endpointUrl")
+            .byDefault()
```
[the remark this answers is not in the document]
catenoid: OK, but this _is_ the way it is written in orika's user guide

Issue #54, opened 2019-03-14
Change phrasing


fdw:
Instead of a "specification" having "versions", it should rather be a "service" having "specifications" (which might differ in their "version" or "language"). As a user, I can better understand these terms because they are more intuitive.

catenoid:
This does slightly conflate with Spring's "@Service" annotation, ie. we end up with 'ServiceFileService' / 'ServiceDataService'

fdw:
To be honest, I don't like the names of these classes: They don't say anything about the class' responsibility. One is probably a `SpecificationDataParser` and the other a `SpecificationFileDownloader`. Since they deal with spec files, they should also keep the word in the name.

Finally,  the `SynchronizationService` should be something like a `RemoteSpecificationUpdater`.

catenoid:
One misgiving I have here is how it interferes with the `/specification/{specificationId}/version/{version}` pattern.

fdw:
That's right. Maybe `/service/{serviceId}/version/{version}`? I think that captures the idea quite well, and since we don't expose the version string as a resource, this is also unambiguous.

PR #67, opened 2019-03-29
Handle uploads to pre-existing versions

### LINKED ISSUE #65: Accept duplicate uploads of the same version
As a developer with a penchant for automatic tooling, I want my ApiCenter-Gradle-Plugin to automatically push the current spec when deploying or releasing so that ApiCenter always has the latest version. However, ApiCenter returns an error code if I try to push a version that already exists even if the content hasn't changed.

I want ApiCenter to compare the content of my new spec against the content of the already uploaded spec with the same version and return a HTTP code 202 (Accepted) if they match. Only if they have actually changed should ApiCenter send a 409 (Conflict) error.

catenoid:
Close #65 

 - When a version upload is attempted, and a database entity with the same ServiceId and version already exists, two potential REST responses are given depending on whether the version contents differ.

 - The interface of the specification repository has changed slightly, replacing `save` by `saveOne` -- which takes a Version, not a Specification, since we only ever want to save one file at a time. `saveOne` performs the file content comparison mentioned above.

Additional changes related to object mappers for Versions:

 - Includes some renaming to makes class names more consistent (ie. not the {Spec -> Service, Version -> Spec} as in issue 54). Goal being that objects associated with aggregating "Versions" are always called "Specifications", which was not always the case previously (eg. replacing "SpecficationMetaData", "SpecificationFileDto").

 - Last attempt of a blanket renaming scheme, this inconsistency caused problems. Everything will be renamed according to #54 next PR.

 - Version objects now contain (in their metadata) which Specification they belong to, which means the object mapper no longer just writes "null" to this foreign key. (This did require adding a new object mapper for SpecificationEntities.)

 - One integration test in the SpecificationController was deleted: in essence it only ever checked that `fromDomain(toDomain(a)) == a`, for which a test already existed in the relevant object mapper unit test.


backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
```diff
@@ -51,5 +51,15 @@ class RestResponseExceptionHandler {
...
                     "is not the same as the API ID used in the specification body (${exception.userDefinedId})", HttpStatus.BAD_REQUEST)
+
+    @ExceptionHandler(PreexistingVersionContentIdenticalException::class)
+    fun handleVersionAlreadyExists(exception: PreexistingVersionContentIdenticalException) =
+            makeResponseEntity("This specification already exists on ApiCenter", HttpStatus.ACCEPTED)
+
+    @ExceptionHandler(PreexistingVersionContentDiscrepancyException::class)
+    fun handleVersionAlreadyExists(exception: PreexistingVersionContentDiscrepancyException) =
+            makeResponseEntity("A specification with the same version name (but different content) already exists." +
+                    "Change the version string. If you're sure no other service depends on this version, consider deleting and reuploading", HttpStatus.CONFLICT)
```
fdw: "There is already a different specification with the same version. Please solve this conflict, for example by increasing the version."

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/SpecificationEntity.kt
```diff
@@ -18,14 +18,14 @@ class SpecificationEntity(
     val remoteAddress: String?
 ) {
 
-    fun pureUpdate(other: SpecificationEntity): SpecificationEntity {
-        val versions = this.versions + other.versions
+    fun appendVersion(other: VersionEntity): SpecificationEntity {
+        val versions = this.versions + other
         return SpecificationEntity(
                 this.id,
                 other.title,
```
fdw: I find it odd that we use sometimes `this` and sometimes `other`. Can we make it as consistent as possible, so it's easy to see where the differences are?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/mapper/converter/SpecificationEntityConverter.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.tngtech.apicenter.backend.domain.entity.ServiceId
+import com.tngtech.apicenter.backend.domain.entity.Specification
+import com.tngtech.apicenter.backend.domain.entity.Version
+import ma.glasnost.orika.MappingContext
+import ma.glasnost.orika.converter.BidirectionalConverter
+import ma.glasnost.orika.metadata.Type
+import org.springframework.stereotype.Component
+
+@Component
+class SpecificationEntityConverter: BidirectionalConverter<Specification, SpecificationEntity>() {
```
fdw: Can't we use a `OrikaMapperFactoryConfigurer` for this?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
```diff
@@ -69,4 +64,39 @@ class SpecificationDatabaseService constructor(
...
+                    version.metadata.title,
+                    version.metadata.description,
+                    listOf(version),
+                    fileUrl
+            )
+            val newSpecificationEntity = specificationEntityMapper.fromDomain(newSpecification)
+            mapAndStoreEntity(newSpecificationEntity)
+        }
+
+        existingSpec.ifPresent { s ->
```
fdw: Can we extract some smaller methods from that?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/dto/VersionFileDto.kt
```diff
@@ -2,18 +2,20 @@ package com.tngtech.apicenter.backend.connector.rest.dto
 
 import com.fasterxml.jackson.annotation.JsonCreator
 import com.tngtech.apicenter.backend.domain.entity.ApiLanguage
+import com.tngtech.apicenter.backend.domain.entity.ServiceId
 
-data class SpecificationMetaData constructor(
+data class VersionMetaData constructor(
```
fdw: I'm not sure what we gain by having this as a separate class. Could you please help me there? :)

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/mapper/VersionFileDtoMapper.kt
```diff
@@ -1,20 +1,19 @@
...
 import com.tngtech.apicenter.backend.connector.rest.dto.VersionDto
+import com.tngtech.apicenter.backend.connector.rest.dto.VersionFileDto
 import com.tngtech.apicenter.backend.domain.entity.Specification
 import com.tngtech.apicenter.backend.domain.entity.Version
 import ma.glasnost.orika.MapperFacade
 import org.springframework.stereotype.Component
 
 @Component
-class VersionDtoMapper constructor(private val mapperFacade: MapperFacade) {
+class VersionFileDtoMapper constructor(private val mapperFacade: MapperFacade) {
```
fdw: Why don't we use Orika directly?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationFileService.kt
```diff
@@ -10,4 +11,13 @@ class SpecificationFileService {
     fun retrieveFile(location: String): String {
         return Scanner(URL(location).openStream(), "UTF-8").useDelimiter("\\A").next()
     }
+
+    fun getLocalOrRemoteFileContent(versionFileDto: VersionFileDto): String {
+        return if (versionFileDto.fileUrl != null && versionFileDto.fileUrl != "") {
```
fdw: I think there should be a `String.isEmpty()` method.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SynchronizationService.kt
```diff
@@ -16,29 +17,18 @@ class SynchronizationService constructor(
 ) {
 
     fun synchronize(specificationId: ServiceId) {
-        val specification = specificationPersistenceService.findOne(specificationId)!!
+        val specification = specificationPersistenceService.findOne(specificationId)
 
-        val remoteAddress = specification.remoteAddress ?: ""
+        if (specification != null) {
```
fdw: How about an "early return", i.e. first testing `specification == null` and throwing the error. Then, without an `else`, you can do the rest of the method.

backend/src/test/kotlin/com/tngtech/apicenter/backend/connector/rest/mapper/converter/VersionFileDtoConverterTest.kt
```diff
@@ -1,93 +1,81 @@
...
-        given(specificationFileService.retrieveFile(SWAGGER_REMOTE)).willReturn(
-            SWAGGER_SPECIFICATION
-        )
-        given(specificationDataService.parseFileContent(SWAGGER_SPECIFICATION)).willReturn(
-            SWAGGER_SPECIFICATION
-        )
+        given(specificationFileService.retrieveFile(SWAGGER_REMOTE)).willReturn(SWAGGER_SPECIFICATION)
+        given(specificationFileService.getLocalOrRemoteFileContent(versionFileDto)).willReturn(SWAGGER_SPECIFICATION)
+
         given(specificationDataService.extractTitle(SWAGGER_SPECIFICATION)).willReturn("Swagger Petstore")
```
fdw: We could collect those in a `@Before` method

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/SpecificationDatabaseService.kt
```diff
@@ -69,4 +64,39 @@ class SpecificationDatabaseService constructor(
...
+            val versionStrings = s.versions.map { v -> versionEntityMapper.toDomain(v).metadata.version }
+            val newVersionString = version.metadata.version
+            if (versionStrings.contains(newVersionString)) {
+
+                val existingContents = s.versions.firstOrNull {
+                    versionEntity -> versionEntityMapper.toDomain(versionEntity).metadata.version == newVersionString
+                }?.content
+
+                if (existingContents == version.content) {
```
fdw: I'd like to do this check in the `domain` package, actually. This is not a characteristic of the DB (or another persistence layer), but of ApiCenter's domain logic.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/SpecificationEntity.kt
catenoid: In fact, I was considering removing title and description from the Specification object. As you see here, these fields are sourced / copied directly from the latest version. So really, this method is always retaining the important details from `this`, ie. the pre-existing specification the new version is to be appended to.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/dto/VersionFileDto.kt
catenoid: The converters to/from the `Version` domain object aren't "symmetric", there's a `VersionDto` (simply content + metadata) and a `VersionFileDto` (from the frontend upload form, whose file content needs to be parsed etc.) Both use the metadata class.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/SpecificationEntity.kt
fdw: Where do suggest we store title and description?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/dto/VersionFileDto.kt
fdw: Yes, that makes sense. Thanks :)

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/SpecificationEntity.kt
catenoid: What I mean is that title / description in the top-level SpecificationEntity object is redundant, if these fields are directly sourced from the latest version contained within the object. I couldn't see a reason for this duplication.

PR #68, opened 2019-03-29
Fix for integer IDs

catenoid:
A bug noticed when attempting to set an `x-api-id` as an integer, the application threw an uncaught exception and the file upload failed. Now the specification upload works as expected.

### DIFF
--- a/backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataService.kt
+++ b/backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataService.kt
@@ -8,6 +8,7 @@ import com.tngtech.apicenter.backend.domain.exceptions.SpecificationParseExcepti
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Service
 import java.io.IOException
+import java.lang.ClassCastException
 
 @Service
 class SpecificationDataService @Autowired constructor(
@@ -74,6 +75,8 @@ class SpecificationDataService @Autowired constructor(
     fun extractId(json: String): String? =
         try {
             JsonPath.read<String>(json, "$.info.x-api-id")
+        } catch (exception: ClassCastException) {
+            JsonPath.read<Integer>(json, "$.info.x-api-id").toString()
         } catch (exception: PathNotFoundException) {
             null
         }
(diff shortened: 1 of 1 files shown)


backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataService.kt
```diff
@@ -74,6 +75,8 @@ class SpecificationDataService @Autowired constructor(
     fun extractId(json: String): String? =
         try {
             JsonPath.read<String>(json, "$.info.x-api-id")
+        } catch (exception: ClassCastException) {
+            JsonPath.read<Integer>(json, "$.info.x-api-id").toString()
```
fdw: If the value is an `Integer`, can't that also be read as a `String`? Also, what happens if it's neither a `String` nor an `Integer`?

catenoid: No, without this catch block, the upload fails. And the type parameter to `JsonPath.read` is required.</br></br>The JSON grammar, being Javascript, only offers "string" and "number" -- However I just tried with a non-integer number, and realised that will also cause an exception / failed uploaded because `java.lang.Double` cannot be cast to either of these

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataService.kt
```diff
@@ -71,10 +72,18 @@ class SpecificationDataService @Autowired constructor(
             null
         }
 
-    fun extractId(json: String): String? =
-        try {
-            JsonPath.read<String>(json, "$.info.x-api-id")
+    fun extractId(json: String): String? {
+        val idPath = "$.info.x-api-id"
+        return try {
+            JsonPath.read<String>(json, idPath)
```
fdw: Do you think we can do it with just Jackson, with `JsonNode.at()` and `JsonNode.asText()`? I think we can avoid the nested `try` with that and also remove another Json dependency ;)

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataService.kt
```diff
@@ -71,10 +71,9 @@ class SpecificationDataService @Autowired constructor(
...
-    fun extractId(json: String): String? =
-        try {
-            JsonPath.read<String>(json, "$.info.x-api-id")
-        } catch (exception: PathNotFoundException) {
-            null
-        }
+    fun extractId(json: String): String? {
+        val parsedJson = objectMapper.readTree(json)
+        val jsonNode = parsedJson.at("/info/x-api-id")
+        return if (jsonNode.isMissingNode) null else jsonNode.asText()
```
fdw: Could you please add braces?

PR #69, opened 2019-04-03
Support view/edit permissions on specifications

### LINKED ISSUE #5: Permission Management
Users should be able to give permissions on specifications they've created. There should be three permission levels:
* view - the user can view the specification
* edit - the user can edit the specification
* admin - the user can give permissions for the specification

Permission settings should be possible by clicking on an icon within the icon bar in the specification overview.

catenoid:
Closes #5 

ACL instantiation

- Adds a permissions manager bean with a simple interface -- for a user and a specification: to add, remove, or check permissions.
- Uses Spring ACL's `BasePermission.READ` / `WRITE`, however within ApiCenter these are renamed to view / edit.
- Configures the ACL datastore according to Spring security's documentation [here](https://docs.spring.io/spring-security/site/docs/3.0.x/reference/domain-acls.html).
- In the current version of Spring security ACL, only resource IDs that are of Java Long type are acceptable. In this PR, if the ID can't be cast to a Long, the ACL features are skipped, making the change backwards compatible for specifications with arbitrary IDs.
- (Some infrastructure exists in Spring security to register a conversion service, to allow for IDs on resources which are not Longs (the goal being Java UUID). However there is a bug making this currently impossible. The fix has been merged to their master branch, but will only be included in the next release, which is not yet available on Maven.)

Filtering based on permissions

- Adds an interceptor to the web security configuration, which acts on all requests with a given specification ID. The interceptor checks the permission for this user and resource based on the HTTP method: GET -> view, POST, PUT, DELETE -> edit
- Filters specifications returned by the SpecificationController's findAll / search by view permission, ie. without view permission, these specifications never reach the frontend.
- For the specifications that *are* returned, if the current user doesn't have edit permission, the edit/delete/edit-permission buttons are not displayed.
- This is, of course, in addition to the interceptor filter in the backend. Trying to forge a url without using the frontend buttons / forms won't succeed. 
- (The button removal is achieved by adding to the Specification DTO a flag indicating if the current principal has edit permission on that Specification.)

Changing permissions on a specification

- Adds a frontend component/page for edit permissions, ie. a new form with a text box (the target username) and two checkboxes (one for view permission, one for edit permission). A successful form submission will change the state of this specification's permissions, for the user specified by the text box, to the permission vector indicated by the checkboxes. To that end...
- Adds a "chmod" API PUT endpoint with pattern `/specification/{specificationId}/chmod/{userId}`, with two 'boolean' (in reality they must be strings) query parameters, being the desired state of the view/edit permissions vector.
- When a permission is deleted, so is its respective permissions record.


backend/build.gradle
```diff
@@ -47,6 +47,7 @@ dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
     compile "org.springframework.boot:spring-boot-starter-web:$spring_boot_version"
     compile "org.springframework.boot:spring-boot-starter-security:$spring_boot_version"
+    compile "org.springframework.security:spring-security-acl:5.1.4.RELEASE"
```
fdw: I think we have another Spring Security dependency. If they use the same versions, can we maybe have a constant for that?

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/AclConfiguration.kt
```diff
@@ -0,0 +1,147 @@
...
+    @Bean
+    fun transactionManager(
+            dataSource: DataSource
+    ): AbstractPlatformTransactionManager {
+        return DataSourceTransactionManager(dataSource)
+    }
+
+    @Bean
+    fun dataSource(): DataSource {
+        val dataSource = DataSourceBuilder
```
fdw: All of this is okay for local tests, but we need something more robust for actual production environments. For example, I can't imagine we will use H2 there. Also, can we move the SQL scripts to an external file and maybe use flyway or something?

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
```diff
@@ -51,5 +51,14 @@ class RestResponseExceptionHandler {
     fun handleServiceIdMismatch(exception: MismatchedSpecificationIdException) =
             makeResponseEntity("The API ID used in the upload URL (${exception.urlPathId}) " +
                     "is not the same as the API ID used in the specification body (${exception.userDefinedId})", HttpStatus.BAD_REQUEST)
+
+    @ExceptionHandler(IdNotCastableToLongException::class)
+    fun handleUncastableId(exception: IdNotCastableToLongException) =
+            makeResponseEntity("A specifications must have an ID that can be parsed to a Java Long type" +
```
fdw: This is going to be fixed with a later Spring Security version, right? Then I'd rather wait until it is fixed before we bake some workarounds into our logic.

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/WebConfiguration.kt
```diff
@@ -1,17 +1,30 @@
...
+import org.springframework.web.servlet.handler.HandlerInterceptorAdapter
+import javax.servlet.http.HttpServletRequest
+import javax.servlet.http.HttpServletResponse
+import java.lang.Long.parseLong
 
 @Configuration
 class WebConfiguration {
 
     @Bean
     fun webMvcConfigurer(): WebMvcConfigurer {
```
fdw: Can't we get the `AclInterceptor` as a dependency by using `fun webMvcConfigurer(aclInterceptor: AclInterceptor)` instead of field wiring?

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/WebConfiguration.kt
```diff
@@ -22,10 +35,50 @@ class WebConfiguration {
...
+        val sid = jwtAuthenticationProvider.getCurrentPrincipal()
+
+        val pathVariables = try {
+            AntPathMatcher("/")
+                    .extractUriTemplateVariables("/api/v1/specifications/{specificationId}/**", request.servletPath)
+        } catch (exception: IllegalStateException) {
+            return true
+        }
+
+        val id = pathVariables["specificationId"]
```
fdw: I can imagine smaller methods ;)

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/acl/service/AclManagerService.kt
```diff
@@ -0,0 +1,121 @@
...
+import org.springframework.transaction.support.TransactionTemplate
+import org.springframework.security.acls.model.MutableAcl
+import org.springframework.security.acls.model.ObjectIdentity
+import mu.KotlinLogging
+import org.springframework.security.acls.domain.BasePermission
+import java.lang.IllegalArgumentException
+
+private val logger = KotlinLogging.logger {  }
+
+interface AclManager {
```
fdw: What do we have the interface for?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/acl/service/AclManagerService.kt
```diff
@@ -0,0 +1,121 @@
...
+        logger.info("Add ${permissionToString(permission)} permission to spec $id ")
+        return aclManagerService.addPermission(Long::class.java, id, sid, permission)
+    }
+
+    fun removePermission(id: Long, sid: Sid, permission: Permission) {
+        logger.info("Remove ${permissionToString(permission)} permission from spec $id ")
+        return aclManagerService.removePermission(Long::class.java, id, sid, permission)
+    }
+
+    fun hasPermission(id: Long, sid: Sid, permission: Permission): Boolean {
```
fdw: To be honest, I don't understand what `sid` is and why it's different from `id`. Could we please rename it?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/acl/service/AclManagerService.kt
```diff
@@ -0,0 +1,121 @@
...
+
+    private fun permissionToString(permission: Permission): String = if (permission == BasePermission.READ) "READ" else "WRITE"
+}
+
+@Service
+class AclManagerService @Autowired constructor(private val aclService: MutableAclService,
+                                               private val transactionManager: AbstractPlatformTransactionManager) : AclManager {
+
+    override fun <T> addPermission(domainClass: Class<T>, id: Serializable, sid: Sid, permission: Permission) {
+        TransactionTemplate(transactionManager).execute(object : TransactionCallbackWithoutResult() {
```
fdw: All this wrapping seems a bit weird. On the other hand, I have no experience with this stuff...

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationController.kt
```diff
@@ -56,9 +76,42 @@ class SpecificationController @Autowired constructor(
         return specificationIdFromPath
     }
 
+    @PutMapping("/{specificationId}/chmod/{userId}")
```
fdw: These are some interesting questions :)
I'd have said that the whole permission thingy should be in the `domain`, because it's not dependent on the interface. On the other hand, the way a user identifies themselves does depend on the interface. So probably some wrapping and converting again and then calling methods in the `domain`?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationController.kt
```diff
@@ -56,9 +76,42 @@ class SpecificationController @Autowired constructor(
...
+
+        } catch (exc: NumberFormatException) {
+            throw IdNotCastableToLongException()
+        }
+
+    }
+
+    private fun changePermission(specificationId: Long, grant: Boolean, sid: PrincipalSid, permission: Permission) {
+        if (grant)
+            permissionManager.addPermission(specificationId, sid, permission)
```
fdw: Important rule: Always use braces! It's more explicit and thus easier to read. And it's harder to make an error...

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/security/JwtAuthenticationProvider.kt
```diff
@@ -24,5 +26,22 @@ class JwtAuthenticationProvider : AuthenticationProvider {
...
+                .build()
+                .verify(authorizationHeader.replace("Bearer ", ""))
+                .subject
+    }
+
+    fun getCurrentUser() =
+            try {
+                (SecurityContextHolder.getContext().authentication as JwtAuthenticationToken).userId
+            } catch (exception: ClassCastException) {
+                "user"
```
fdw: Shouldn't this result in an exception? Under which circumstances is there no token?

fdw:
@catenoid : Apparently, Spring Security 5.2.0 will be out in September. So I think we should rather finish this thing up and merge it, even if it's not perfect.

catenoid:
Superseded by #179

PR #71, opened 2019-04-04
Compare JWT subject in frontend authentication

catenoid:
Previously, the only check that Angular's authentication guard provided was that the string in `window.localStorage.token` was not null.

If a user sets the token to any random string (with the browser console), they bypass the guard, and can get to `/`. No specifications appear in the table, because the backend will reject the 'token' and so respond with Spring's 403 default response. However no error message is displayed; it simply looks as though no specifications have ever been uploaded.

This PR makes the guard _slightly_ more sophisticated. It decodes the current token, and checks the token's subject matches the username in local storage. Failure will redirect to `/login` again.

We cannot, client-side, verify that the token is signed correctly, because the HMAC secret stays server-side. If a user forges a token on [jwt.io](https://jwt.io/), they can get around the frontend guard. (This is equivalent to -- before the changes in this PR -- setting the token to a random string, as described above.)

jwt-io

console

However now there is an error displayed when the homepage receives a 403 response from the backend.

permission-error

### DIFF
--- a/backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/security/JwtAuthorizationFilter.kt
+++ b/backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/security/JwtAuthorizationFilter.kt
@@ -2,6 +2,7 @@ package com.tngtech.apicenter.backend.connector.rest.security
 
 import com.auth0.jwt.JWT
 import com.auth0.jwt.algorithms.Algorithm
+import com.auth0.jwt.exceptions.SignatureVerificationException
 import org.springframework.security.authentication.AuthenticationManager
 import org.springframework.security.core.context.SecurityContextHolder
 import org.springframework.security.web.authentication.www.BasicAuthenticationFilter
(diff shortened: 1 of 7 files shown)


backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/security/JwtAuthorizationFilter.kt
```diff
@@ -23,13 +24,17 @@ class JwtAuthorizationFilter(authManager: AuthenticationManager, private val jwt
...
+                    .build()
+                    .verify(header.replace("Bearer ", ""))
+                    .subject
 
-        SecurityContextHolder.getContext().authentication =
-            if (user != null) JwtAuthenticationToken(user, header) else null
+            SecurityContextHolder.getContext().authentication =
+                    if (user != null) JwtAuthenticationToken(user, header) else null
+        } catch (exception: SignatureVerificationException) {
+            // Token's signature invalid when verified using server-side secret and HMAC512"
```
fdw: Shouldn't we do something here? What happens if the `SignatureVerificationException` is caught?

catenoid: The catch is not strictly needed, but maybe a log message could go there? In the case of a badly signed token, the filter mirrors the behaviour, as when the Authorization header is absent, namely `chain.doFilter(req, res)` then `return`, user authentication will fail as expected.

fdw: Okay, thanks :)

PR #73, opened 2019-04-15
Fix for file uploads with metadata

catenoid:
#68 changed the `extractId` function to accept multiple data types, but accidentally removed the case when the file could not be parsed as JSON, and hence, no ID was extractable.

The try / catch block is reintroduced to avoid an uncaught exception causing the upload to fail.
***
#67 made the specification ID a required field in a version's metadata.

When uploading a GraphQL file, the user specifies certain fields of the metadata explicitly (namely title and version, because they cannot be derived from the file contents). But not the specification's ID, this is assigned server-side to a random UUID.

Because the metadata object used for file uploads (no ID) was not the same as the metadata object used in the domain layer (has ID, and cannot be null), Spring raised an `HttpMessageNotReadable` exception when casting the `@RequestBody` that the ID field was missing.

Now there are two different metadata objects defined; there is, regrettably, a fair amount of overlap between them. Hopefully there is a better way of doing this -- My reasoning for each sub-object was along the lines of: `VersionFileMetaData` is useful because the whole object should be optional, and `VersionMetaData` is useful because it means the `makeSpecificationMetaData` function be extracted.
***
Adds a test which would have detected these problems 😓

### DIFF
--- a/backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceControllerIntegrationTest.kt
+++ b/backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceControllerIntegrationTest.kt
@@ -79,6 +79,23 @@ internal class ServiceControllerIntegrationTest {
     }
 
     @Test
+    fun uploadSpecification_shouldCreateSpecificationWithMetadata() {
+        mockMvc.perform(
+                post("/api/v1/service")
+                        .with(user("user"))
+                        .with(csrf())
+                        .contentType("application/json")
+                        .content("""
+                           {"fileContent":"t","metadata":{"title":"My title","version":"v3","description":"","language":"GRAPHQL","endpointUrl":""}}
+                        """.trimIndent()
+                        )
+        )
+                .andExpect(status().isCreated)
+                .andExpect(jsonPath("$.metadata.title", equalTo("My title")))
+                .andExpect(jsonPath("$.metadata.version", equalTo("v3")))
+    }
+
+    @Test
     fun uploadSpecification_shouldDetectVersionClash_identicalContent() {
         mockMvc.perform(
                 post("/api/v1/service")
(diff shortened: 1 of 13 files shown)


backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataService.kt
```diff
@@ -74,24 +77,52 @@ class SpecificationDataService @Autowired constructor(
...
+            if (jsonNode.isMissingNode) {
+                null
+            } else {
+                jsonNode.asText()
+            }
+        } catch (exception: IOException) {
             null
+        }
+
+    fun makeSpecificationMetaData(fileContent: String,
```
fdw: If I understand that correctly, the decision whether it is `GRAPHQL` or `OPENAPI` hinges on whether there is `VersionFileMetaData`. To be honest, I find that a bit opaque and unexpected. Do you think there is a clearer way?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataService.kt
```diff
@@ -74,24 +77,52 @@ class SpecificationDataService @Autowired constructor(
...
-        return VersionMetaData(
-                serviceId,
-                extractTitle(fileContent),
-                extractVersion(fileContent),
-                extractDescription(fileContent),
-                ApiLanguage.OPENAPI,
-                endpointUrl
-        )
+    private fun getConsistentServiceId(idFromUpload: String?, idFromPath: String?): ServiceId {
+        if (idFromUpload != null && idFromPath != null && idFromUpload != idFromPath) {
```
fdw: `Objects.equals` should be able to do something very similar

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataService.kt
catenoid: I completely agree, and have been thinking about changing it. However maybe instead the language enum itself here is misleading. In my mind, the distinction ApiCenter cares about (when uploading) is whether or not the metadata can be sourced from the file contents, or needs to be additionally supplied by the user uploading the file --> ie. whether or not the frontend sends metadata. This could apply to other formats which do not include metadata required by ApiCenter in their specifications. The enum was initially added to simply choose between the SwaggerUI and GraphiQL. Perhaps adding a map from ApiLanguage values to whether they need metadata would make the distinction clear.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataService.kt
catenoid: When both are null, the expression should return false (no exception thrown), whereas `Object.equals` would return true.

PR #75, opened 2019-04-16
Rename Specification -> Service; Version -> Specification

### LINKED ISSUE #54: Change phrasing

catenoid:
Close #54


frontend/src/app/service-store.service.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  const httpClient = instance(mockedHttpClient);
+
+  const metadataStub = {
+    version: '1.0.0',
+    description: 'Description',
+    language: ApiLanguage.OpenAPI,
+    endpointUrl: null,
+  };
+
+  const firstVersions = [new Specification('Content', {...metadataStub, title: 'API 1'})];
```
fdw: I think the variable names would like to be changed as well :p

backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceControllerIntegrationTest.kt
```diff
@@ -0,0 +1,243 @@
...
+@SpringBootTest
+@AutoConfigureMockMvc
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+internal class ServiceControllerIntegrationTest {
+
+    @Autowired
+    private lateinit var mockMvc: MockMvc
+
+    @Test
+    fun findAllSpecifications_shouldReturnAllSpecifications() {
```
fdw: I don't think all of these method names make sense anymore

backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
```diff
@@ -8,234 +8,63 @@ import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMock
...
-                                """
-                            | {
-                            |   "id": "b6b06513-d259-4faf-b34b-a216b3daad6a",
-                            |   "fileContent": "{\"info\": {\"title\": \"Spec1\",  \"version\": \"v1\", \"description\": \"I'm different\"}}"
-                            | }
-                            """.trimMargin()
-                        )
-        )
-                .andExpect(status().isConflict)
+    fun findOneVersion_shouldReturnVersion() {
```
fdw: Same here (method names)

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/crowd/service/CrowdAuthentication.kt
```diff
@@ -2,15 +2,15 @@ package com.tngtech.apicenter.backend.connector.crowd.service
...
+import com.tngtech.apicenter.backend.domain.service.ExternalAuthentication
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.stereotype.Service
 import java.util.UUID
 
 @Service
 @ConditionalOnProperty(name = ["auth.service"], havingValue = "crowd")
-class CrowdAuthenticationService @Autowired constructor(private val crowdClient: CrowdClient) : ExternalAuthenticationService {
+class CrowdAuthentication @Autowired constructor(private val crowdClient: CrowdClient) : ExternalAuthentication {
```
fdw: I'm for ditching "Service", but - at least in my opinion - class names should describe what they're doing (e.g. "Authenticator"). More "passive" names are reserved for data classes, value objects etc.
Does that make sense?

catenoid: ie. `*Persistence` to `*Persister`?

fdw: Yes. Kind of. "Persister" feels unnatural, but it also feels awkward to call `Persistence.find()`. Maybe "repository", or "archive"? (Or even "vault"? ;))

fdw: And I just noticed that "archive" and "vault" aren't that active, either. I think concreteness also plays a role here, as I can better imagine what an "archive" does than what "persistence" does.

As an aside, "persistence" would make - in my opinion - for a good package name, as these tend to describe the common theme, but nothing concrete like a class.

PR #79, opened 2019-04-26
Update Swagger file after renaming

catenoid:
Adds tags which were undefined. Removes status code responses which aren't served. Corrects specifying the required schema fields as a list and not in the properties section of the object.

Updates some parameters in the ServiceController that were overlooked previously.

### DIFF
--- a/backend/src/main/resources/api-definition.json
+++ b/backend/src/main/resources/api-definition.json
@@ -17,7 +17,7 @@
     },
     {
       "name": "specification-controller",
-      "description": "Operations to find, delete individual specifications"
+      "description": "Operations to find and delete individual specifications"
     },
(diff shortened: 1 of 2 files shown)


backend/src/main/resources/api-definition.json
```diff
@@ -11,21 +11,29 @@
...
   "basePath": "/",
   "tags": [
+    {
+      "name": "service-controller",
+      "description": "Operations to create, update, delete, and search for services"
+    },
     {
       "name": "specification-controller",
-      "description": "Operations to create, update and maintain specifications"
+      "description": "Operations to find, delete individual specifications"
```
fdw: "and"?

PR #80, opened 2019-05-08
Add configuration for docker

catenoid:
Workflow is:

- `./gradlew :buildDockerImage` or `docker pull catenoid/apicenter:extdb` to create/get the image
- `docker-compose up`
- ...
- `docker-compose down` (add `-v` to clear the volume containing the application data.)


build.gradle
```diff
@@ -3,7 +3,13 @@ task monoBuild {
     dependsOn(':frontend:buildClientDev', ':frontend:copyFrontendFiles', ':backend:bootRun')
 }
 
+task buildDockerImage {
+    description = 'Build Docker image'
+    dependsOn(':frontend:buildClientDev', ':frontend:copyFrontendFiles', ':backend:dockerBuildImage')
+}
+
 tasks.getByPath(':backend:bootRun').mustRunAfter(':frontend:copyFrontendFiles')
+tasks.getByPath(':backend:dockerBuildImage').mustRunAfter(':frontend:copyFrontendFiles')
```
fdw: Do we need this line if it's already in the `dependsOn` in line 8?

backend/build.gradle
```diff
@@ -27,6 +27,9 @@ plugins {
     id "com.github.ben-manes.versions" version "0.20.0"
     id "com.diffplug.gradle.spotless" version "3.14.0"
     id "info.solidsoft.pitest" version "1.3.0"
+    id "com.moowork.node" version "0.11"
```
fdw: As always: Have you checked the licences for the new dependencies? Are their terms permissive enough for us?

backend/build.gradle
```diff
@@ -94,3 +97,10 @@ task integrationTest(type: Test) {
     classpath = sourceSets.integrationTest.runtimeClasspath
     outputs.upToDateWhen { false }
 }
+
+docker {
+    springBootApplication {
+        baseImage = 'openjdk:8-alpine'
```
fdw: What are the terms on that? Are we allowed to build derivative works?

build.gradle
catenoid: Line 8 adds those tasks as dependencies, but does not specify any mutual ordering between them. And `mustRunAfter` alone does not imply execution dependency, the rule only applies if both tasks are scheduled. ([docs](https://docs.gradle.org/current/userguide/more_about_tasks.html))

backend/build.gradle
catenoid: All 3 are Apache-2.0 licensed, permitting reuse.

backend/build.gradle
catenoid: OpenJDK is GPLv2 licensed... 🤔

catenoid: The Docker base image used for the Spring Boot application is [here](https://github.com/docker-library/openjdk) MIT licensed. From [here](https://law.stackexchange.com/questions/6081/can-i-bundle-mit-licensed-components-in-a-apache-2-0-licensed-project) both are permissive, so we can relicense our derivative work however we want (ie. with Apache-2.0).

backend/build.gradle
catenoid: [Source](https://github.com/srs/gradle-node-plugin)

backend/build.gradle
fdw: Thanks for checking them all :)

PR #82, opened 2019-05-20
Fix Heroku deploy

catenoid:
Previously you could view and login to ApiCenter, but uploads would fail, returning from the backend an 500 Internal Server Error, root cause `org.postgresql.util.PSQLException: ERROR: column specificat1_.service_id does not exist`.

Assuming the column names in the code (in camelCase) were being converted inconsistently to snake_case, found a solution [here](https://stackoverflow.com/questions/25283198/spring-boot-jpa-column-name-annotation-ignored/38875123#38875123).

### DIFF
--- a/backend/src/main/resources/heroku.properties
+++ b/backend/src/main/resources/heroku.properties
@@ -2,3 +2,4 @@ jwt.secret=${JWT_SECRET}
 spring.datasource.url=${DATABASE_URL}
 spring.datasource.driver-class-name=org.postgresql.Driver
 spring.jpa.generate-ddl=true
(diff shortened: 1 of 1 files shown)


backend/src/main/resources/heroku.properties
```diff
@@ -1,4 +1,5 @@
 jwt.secret=${JWT_SECRET}
 spring.datasource.url=${DATABASE_URL}
 spring.datasource.driver-class-name=org.postgresql.Driver
-spring.jpa.generate-ddl=true
\ No newline at end of file
+spring.jpa.generate-ddl=true
+spring.jpa.hibernate.naming.physical-strategy=org.hibernate.boot.model.naming.PhysicalNamingStrategyStandardImpl
```
fdw: Ahh! That is a nice find! :)

PR #103, opened 2019-05-28
Make README friendlier

catenoid:
Attempt to move the more technical / development-oriented content from the README to the wiki, and add information on how to run the docker image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,13 +4,16 @@
 
 ApiCenter is a repository for OpenAPI and GraphQL specifications. Through the web interface you can upload and download specifications, interact with them through the [SwaggerUI](https://swagger.io/tools/swagger-ui/) / [GraphiQL](https://github.com/graphql/graphiql), and search their contents.
 #### Specification overview
-![Specification overview screenshot](https://user-images.githubusercontent.com/6897716/58478901-651c4480-8157-11e9-90d1-b4c3957c64d3.PNG)
+![Specification overview screenshot](https://user-images.githubusercontent.com/6897716/59666928-1beb6d80-91b6-11e9-9cf6-f2fdb788a903.PNG)
 
 #### SwaggerUI view 
 
-![SwaggerUI view](https://user-images.githubusercontent.com/6897716/58482301-6866fe80-815e-11e9-9c76-3e748bb610fe.PNG)
+![SwaggerUI view](https://user-images.githubusercontent.com/6897716/59666922-18f07d00-91b6-11e9-90b8-5a4d23d0adad.PNG)
 
 More information and troubleshooting is on the [wiki](https://github.com/TNG/ApiCenter/wiki).
+
+#### Demo file upload
+![Demo file upload](https://user-images.githubusercontent.com/6897716/59666747-bdbe8a80-91b5-11e9-8676-eaccb12e24d3.gif)
 
 ## How to run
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,106 +2,36 @@
 
 # ApiCenter
 
-ApiCenter is a repository for your OpenAPI and GraphQL specifications. Through the web interface you can upload specifications, interact with them through the [SwaggerUI](https://swagger.io/tools/swagger-ui/) / [GraphiQL](https://github.com/graphql/graphiql), and search their contents.
+ApiCenter is a repository for OpenAPI and GraphQL specifications. Through the web interface you can upload & download specifications, interact with them through the [SwaggerUI](https://swagger.io/tools/swagger-ui/) / [GraphiQL](https://github.com/graphql/graphiql), and search their contents.
```
fdw: "and"

README.md
```diff
@@ -2,106 +2,36 @@
...
-#### Combined
-From the project root directory, run:
-```
-./gradlew :monoBuild
-```
-The `:monoBuild` task is composed of three subtasks, which can be independently invoked if needed:
-- `:frontend:buildClientDev` runs Angular's build command, producing the output files in the default location `frontend/dist`.
-- `:frontend:copyFrontendFiles` copies these files into a directory in `backend` from which Spring will serve static resources.
-- `:backend:bootRun` will compile and run the backend.
+* In a browser, go to `<that-IP-Address>:8080`
```
fdw: Isn't there something possible with `localhost:8080`?

README.md
```diff
@@ -2,106 +2,36 @@
...
 
-The service is available from `localhost:8080` (Spring boot's default port).
+By default, user authentication is disabled. Login succeeds regardless of what password is entered.
 
-#### Separate
+## How to test
```
fdw: I'd rather move this to the wiki

README.md
```diff
@@ -2,106 +2,36 @@
 
 # ApiCenter
 
-ApiCenter is a repository for your OpenAPI and GraphQL specifications. Through the web interface you can upload specifications, interact with them through the [SwaggerUI](https://swagger.io/tools/swagger-ui/) / [GraphiQL](https://github.com/graphql/graphiql), and search their contents.
+ApiCenter is a repository for OpenAPI and GraphQL specifications. Through the web interface you can upload & download specifications, interact with them through the [SwaggerUI](https://swagger.io/tools/swagger-ui/) / [GraphiQL](https://github.com/graphql/graphiql), and search their contents.
 
-## How to use
+More information and troubleshooting is on the [wiki](https://github.com/TNG/ApiCenter/wiki).
```
fdw: Some screenshots, maybe?

catenoid: I decided against this, thinking that for the README to render locally, these images would have to be committed as part of the project. And the screenshots aren't really static assets, they'd need updating in response to cosmetic UI changes, which in my mind would just create noise in the commit history. It's much simpler to manage screenshots / related media on the wiki, as Github can host everything. Does that seem reasonable? I am open to counter arguments.

fdw: Your argument definitely makes sense. My only counterpoint is that the readme is the first things people see, as it is shown on the main Github page. If we don't capture interest here, they won't click through to the wiki.

Also, the static web page would require screenshots, too, and it is usually also kept in the repo (see Github pages).

PR #136, opened 2019-06-18
Add flyway init script

catenoid:


fdw:
If I understand the whole magic correctly, the actual table definitions come from Hibernate/Spring JPA, right?

PR #149, opened 2019-06-24
Spec overview table formatting

### LINKED ISSUE #104: Make column spacing consistent when first spec added
The column spacing changes slightly between the state of having no specifications, and >=1 specification(s). It's visually a bit jarring. These were taken with the same browser viewport before and after.

no-specs

one-spec

catenoid:
Close #104, purely cosmetic.

Add explicit column widths and rotate chevron on click

### DIFF
--- a/frontend/src/app/specification-overview/specification-overview.component.css
+++ b/frontend/src/app/specification-overview/specification-overview.component.css
@@ -2,12 +2,16 @@ button {
   margin-right: 5px;
 }
 
+#chevron-column {
+  width: 5%;
+}
+
 #title-column {
   width: 60%;
 }
 
 #version-column {
-  width: 20%;
+  width: 15%;
 }
 
 #actions-column {
--- a/frontend/src/app/specification-overview/specification-overview.component.css
+++ b/frontend/src/app/specification-overview/specification-overview.component.css
@@ -1,3 +1,16 @@
 button {
   margin-right: 5px;
 }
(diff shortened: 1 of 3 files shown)


frontend/src/app/specification-overview/specification-overview.component.css
```diff
@@ -1,3 +1,16 @@
...
+#title-column {
+  width: 60%;
+}
+
+#version-column {
+  width: 20%;
+}
+
+#actions-column {
+  width: 20%;
```
fdw: Can we tell it to make this as small as possible, and use the rest for the other columns?

fdw:
This sounds useful, too: https://stackoverflow.com/a/15884087

frontend/src/app/specification-overview/specification-overview.component.css
catenoid: I'm a little uneasy about mixing up pixel and percent measurements. (which I assume was the proposition, as the oi buttons are of fixed size?) After searching I couldn't find a clear consensus as to whether it was safe. Also, the button to edit permissions on a specification conditionally appears based on whether the logged in user has those access rights, so I did not see this column as having an objective minimum width.

fdw: I'd probably have tried to set it to some ridiculous amount like `5%` :)

PR #150, opened 2019-06-24
Small fixes to search functionality

catenoid:
Found when testing

Details:
- Corrects a method name that was accidentally skipped in the grand `Specification` -> `Service` renaming
- The test `services.length == 0` was occurring before the services list was defined, resulting in a frontend console error. Now there's a boolean to guard this field before the data comes through
- Pressing enter in an empty search box will not update the results or refresh the page
- If the user entered only characters that are filtered out by Spring's controllers (eg. just `#`, or ` `), so that the `@PathVariable` would end up blank, the route requested of the backend was `/api/v1/search/`. This is not matched by any controller, resulting in an uncaught server error. (Open to alternative suggestions on how to solve this than the one in this PR)

### DIFF
--- a/backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceController.kt
+++ b/backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceController.kt
@@ -83,4 +83,7 @@ class ServiceController @Autowired constructor(
     @GetMapping("/search/{searchString}")
     fun searchForServices(@PathVariable searchString: String): List<ServiceDto> =
         serviceHandler.search(searchString).map { spec -> serviceDtoMapper.fromDomain(spec) }
+
+    @GetMapping("/search/")
+    fun searchWithBlank(): List<ServiceDto> = listOf()
 }
--- a/frontend/src/app/specification-search-detail/specification-search-detail.component.ts
+++ b/frontend/src/app/specification-search-detail/specification-search-detail.component.ts
@@ -29,7 +29,7 @@ export class SpecificationSearchDetailComponent implements OnInit {
   }
 
   public async searchServices(event) {
-    if (event.keyCode === 13 && this.searchString != "") {
+    if (event.keyCode === 13 && this.searchString !== undefined && this.searchString !== '') {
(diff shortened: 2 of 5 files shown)


backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceController.kt
```diff
@@ -83,4 +83,7 @@ class ServiceController @Autowired constructor(
     @GetMapping("/search/{searchString}")
     fun searchForServices(@PathVariable searchString: String): List<ServiceDto> =
         serviceHandler.search(searchString).map { spec -> serviceDtoMapper.fromDomain(spec) }
+
+    @GetMapping("/search/")
+    fun searchWithBlank(): List<ServiceDto> = listOf()
```
fdw: Interesting that this isn't caught by the previous method.

frontend/src/app/specification-search-detail/specification-search-detail.component.ts
```diff
@@ -12,21 +12,27 @@ import {ActivatedRoute} from '@angular/router';
...
+          this.searchComplete = true;
+        });
       this.searchString = params['searchString'];
     });
   }
 
   public async searchServices(event) {
-    if (event.keyCode === 13) {
-      this.serviceStore.searchForServices(this.searchString).subscribe((data: Service[]) => this.services = data);
+    if (event.keyCode === 13 && this.searchString !== undefined && this.searchString !== '') {
```
fdw: Should we show a warning for the other cases? So that the user doesn't wonder why hitting enter does not work.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceController.kt
catenoid: It makes sense to me when thinking of a rest API that returns a collection of all entries for `/entries` and a specific entry for `/entries/{uuid}`, for example. I believe ApiCenter implements something similar. This workaround makes me uneasy because it is logically an exception case of the GetMapping above, and the code doesn't reflect that. But I'm not sure if it's clearer to do something like this https://stackoverflow.com/a/34670841

PR #154, opened 2019-06-26
Update to Angular 7

fdw:
This tries to update a lot of stuff as far as I can make it.
Unfortunately, ng-bootstrap is not yet ready for Angular 8, so we have
to use 7.


fdw:
Should be better now, @catenoid :)

PR #155, opened 2019-06-27
Display endpoint changes

### LINKED ISSUE #57: Add changelogs/diffs to versions
As a developer of microservices, it would be very comfortable if I could easily compare different versions of one service's API. They could either be automatic diffs of the spec or, better, manually created changelogs between versions that the service's maintainers would have to create.

catenoid:
Close #57 

Demo:
diffs

This uses [OpenAPI-diff](https://github.com/quen2404/openapi-diff) to interpret the endpoint changes from the OpenAPI files, and an ng-bootstrap Tabset to display the tables.

I could add an additional string metadata field to serve as a changelog when uploading a specification, which the uploader fills with a human readable summary of the changes / release notes. Idea being this would be displayed alongside the tables. Would this be useful?


catenoid:
NB openapi-diff only supports swagger 3.x

frontend/src/app/diff-versions/diff-versions.component.html
```diff
@@ -0,0 +1,121 @@
...
+          <th scope="col">-</th>
+        </tr>
+        </thead>
+        <ng-container *ngFor="let specification of (searchComplete ? service.specifications : [])">
+          <tbody *ngIf="service.specifications.length > 1">
+          <tr>
+            <td><a [routerLink]="['specifications/', service.id, specification.metadata.version]" routerLinkActive="active">{{ service.title }}</a></td>
+            <td>{{ specification.metadata.version }}</td>
+            <td>
+              <input type="radio" name="radiogroup-before" [value]="specification.metadata.version" (change)="radioButtonChange($event)" [(ngModel)]="radioSelectedBeforeVersion"/>
```
fdw: I'm not sure if this too complicated. I always hate that config dialog where I have to choose two versions to compare. What do you think about automatically comparing it with the previous version?

catenoid: Just to clarify, are you proposing a) pre-setting the radio buttons to versions latest and (latest-1), b) removing the radio button selection entirely and only ever displaying the results of latest vs (latest-1), or c) only having one column of radio buttons, which displays the diff between version n and version (n-1)?

fdw: My idea was b). However, that is just an idea I'd like you input on :)

catenoid: I think ultimately it's useful to have the flexibility of being about to choose any pair, for example to compare across major versions. (As a compromise, I could have multiple views as part of the "diff-versions" Angular page, where the user first sees the results of b), however can navigate to a tab with a more powerful but complex interface?)

catenoid: Ideally I wanted to UI element to be a vertical double-ended range slider, sort of like https://codepen.io/Gyuszko/pen/QrJoEz However I could not find a way to integrate this into an html table, as the table rows correspond to the slider options.

PR #156, opened 2019-06-27
Prereleases and snapshot versions

### LINKED ISSUE #63: Support prerelease versions
Sometimes I, as a developer, want to "pre-publish" an in-development version of a spec to show the current changes or to test it, similar to artifact releases. This will help me to discuss changes with developers of other services, for example. 

Such in-development versions must be appended with "-BETA[\d]" or "-RC[\d]", for example "2.1.0-BETA1" or "2.1.0.RC3".

Differences to the stable versions:
- The main version of a spec should be the latest non-snapshot. Only if you unfold it, you can see the snapshot-versions.

### LINKED ISSUE #64: Support SNAPSHOT-versions
As a developer, I don't want a specific spec version of a service to change, so that communication is easier and clearer. However, there should be support for "SNAPSHOT" versions that are not stable and are allowed to change. I can use these for the development of the next spec version and share them even in their non-final state.

Spec versions can be defined as a snapshot by appending "-SNAPSHOT" to the version, for example "4.1.3-SNAPSHOT".

Snapshots are generally considered pre-release versions and should behave as such (see #63). Additionally, the have the following properties:
- They are not immutable and a different spec with the same version can be uploaded. It overwrites the previous content.

catenoid:
Close #63 and close #64. Grouped these, as they are closely related when considering changes to the specification data model.

- The main version is that latest which is non-snapshot and non-prerelease, and only visible after expanding a table node in the overview
- When uploading a snapshot version, permit overriding of the specification content

snapshots

### DIFF
--- a/backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/entity/Specification.kt
+++ b/backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/entity/Specification.kt
@@ -3,19 +3,18 @@ package com.tngtech.apicenter.backend.domain.entity
 import java.util.regex.Pattern
 
 enum class ReleaseType {
-    RELEASE, PRERELEASE, SNAPSHOT
-}
+    RELEASE, PRERELEASE, SNAPSHOT;
 
-fun inferReleaseType(version: String): ReleaseType {
-    return if (version.endsWith("-SNAPSHOT")) {
-        ReleaseType.SNAPSHOT
-    } else if (Pattern.matches("-BETA\\d*", version) || Pattern.matches("-RC\\d*", version)) {
-        ReleaseType.PRERELEASE
-    } else {
-        ReleaseType.RELEASE
+    companion object {
+        fun fromVersionString(version: String): ReleaseType {
+            return when {
+                version.endsWith("-SNAPSHOT") -> SNAPSHOT
+                Pattern.matches("-(BETA|RC)\\d*", version) -> PRERELEASE
+                else -> RELEASE
+            }
+        }
     }
 }
-
 
 data class SpecificationMetadata constructor(
     val id: ServiceId,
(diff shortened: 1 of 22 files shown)


frontend/src/app/models/specification.ts
```diff
@@ -5,6 +5,22 @@ export enum ApiLanguage {
   GraphQL = 'GRAPHQL',
 }
 
+export enum ReleaseType {
+  Release = 'RELEASE',
+  Prerelease = 'PRERELEASE',
+  Snapshot = 'SNAPSHOT',
+}
+
+export function inferReleaseType(version: string): ReleaseType {
```
fdw: Can't we send this information from the backend isntead of inferring it in the frontend?

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/entity/Specification.kt
```diff
@@ -1,11 +1,29 @@
 package com.tngtech.apicenter.backend.domain.entity
 
+import java.util.regex.Pattern
+
+enum class ReleaseType {
+    RELEASE, PRERELEASE, SNAPSHOT
+}
+
+fun inferReleaseType(version: String): ReleaseType {
+    return if (version.endsWith("-SNAPSHOT")) {
```
fdw: First, I'd like those patterns in a constant (you should be able to do `-(BETA|RC)` for the `PRERELEASE`.
Secondly, I think this should be a `fromVersionString` method in the `enum class` itself.

catenoid: To the first point (after belatedly actually reading the semver spec...), it seems we ought to be a bit more liberal in the pre-release acceptance criteria? https://semver.org/#spec-item-9 (Or would you rather only have these extensions?)

fdw: I can live with both. On the one hand, supporting the full Semver spec would be nice, but on the other hand, one should not have different versions of one version...

So whatever is easier, I'd say

PR #172, opened 2019-07-01
Support multiple files in specification form upload

catenoid:
Allow the user to upload multiple specifications in a single form submission

### DIFF
--- a/backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceControllerIntegrationTest.kt
+++ b/backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceControllerIntegrationTest.kt
@@ -67,15 +67,15 @@ internal class ServiceControllerIntegrationTest {
                 .contentType("application/json")
                 .content(
                     """
-                            | {
+                            | [{
                             |   "fileContent": "{\"info\": {\"title\": \"Spec\",\"version\": \"1.0.0\"}}"
-                            | }
+                            | }]
                             """.trimMargin()
                 )
         )
             .andExpect(status().isCreated)
-            .andExpect(jsonPath("$.metadata.title", equalTo("Spec")))
-            .andExpect(jsonPath("$.metadata.version", equalTo("1.0.0")))
+            .andExpect(jsonPath("$[0].metadata.title", equalTo("Spec")))
+            .andExpect(jsonPath("$[0].metadata.version", equalTo("1.0.0")))
     }
 
     @Test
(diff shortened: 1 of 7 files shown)


backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceControllerIntegrationTest.kt
```diff
@@ -67,15 +67,15 @@ internal class ServiceControllerIntegrationTest {
...
                             |   "fileContent": "{\"info\": {\"title\": \"Spec\",\"version\": \"1.0.0\"}}"
-                            | }
+                            | }]
                             """.trimMargin()
                 )
         )
             .andExpect(status().isCreated)
-            .andExpect(jsonPath("$.metadata.title", equalTo("Spec")))
-            .andExpect(jsonPath("$.metadata.version", equalTo("1.0.0")))
+            .andExpect(jsonPath("$[0].metadata.title", equalTo("Spec")))
```
fdw: and check the size?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceController.kt
```diff
@@ -25,12 +29,12 @@ class ServiceController @Autowired constructor(
 
     @PostMapping
     @ResponseStatus(HttpStatus.CREATED)
-    fun uploadSpecification(@RequestBody specificationFileDto: SpecificationFileDto): SpecificationDto {
-        val specification = specificationFileDtoMapper.toDomain(specificationFileDto)
-
-        serviceHandler.addNewSpecification(specification, specification.metadata.id, specificationFileDto.fileUrl)
-
-        return specificationFileDtoMapper.fromDomain(specification)
+    fun uploadSpecifications(@RequestBody specificationFileDtos: List<SpecificationFileDto>): List<Specification> {
```
fdw: Is this a `List` or a `Set`?

frontend/src/app/specification-form/specification-form.component.html
```diff
@@ -13,10 +13,10 @@ <h4 class="modal-title">Local file upload</h4>
        <div class="alert alert-danger" role="alert" *ngIf="error">
          {{ error }}
        </div>
-       <input type="file" placeholder="Upload Swagger Specification" accept=".yml,.json,.yaml,.graphql" ngModel
+       <input multiple type="file" accept=".yml,.json,.yaml,.graphql" ngModel
```
fdw: I think I'd still like a placeholder, if possible

frontend/src/app/specification-form/specification-form.component.ts
```diff
@@ -97,38 +97,81 @@ export class SpecificationFormComponent implements OnInit {
...
       this.error = 'No file selected';
       return;
     }
 
-    const reader = new FileReader();
-    const me = this;
+    const mutableFileList = Array.from(this.specificationFiles);
+    if (this.specificationFiles.length > 1 &&
+      mutableFileList.some(file => /.*\.graphql/.test(file.name))) {
+      this.error = 'Multi-upload is only allowed when every file includes all the required metadata (ie. are all of the OpenAPI format)';
```
fdw: I think we should simplify the message to only include the "i.e." part

catenoid:
> 
> 
> The only thing I wonder about this change is how often someone would use it. I'd expect people to only work on the API of one service at the same time, so I'm don't know when they would need to upload several.

@fdw Granted, probably only useful the first time someone sets-up the application and has multiple historic services to upload. But in that case, if you've any more than 3 APIs, it saves a noticeable amount of clicking (and also simplifies manual testing). Also, it's a feature I would just expect to be supported somehow...

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceController.kt
catenoid: It's a list, my rational being it's fundamentally receiving JSON, which only supports ordered arrays and keyed objects. (This signature is how it appears in the documentation.)

frontend/src/app/specification-form/specification-form.component.html
catenoid: This placeholder never actually appeared 😄 The placeholder attribute is only used for text input types

PR #173, opened 2019-07-02
Exception for malformed remote URL

catenoid:
Submitting something like "hello" to the remote file form causes an exception that wasn't part of the rest response exception handler, and so did not return an error message in a way the frontend could use:

`{"timestamp":1562056004073,"status":500,"error":"Internal Server Error","message":"no protocol: hello","path":"/api/v1/service"}`

This change also removes the string the user gave (ie. `location`) from the error message, because it already appears directly below in the form text box. (However it still might be useful for backend logging, so it's been left in the Exception constructors.)

### DIFF
--- a/backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
+++ b/backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
@@ -63,6 +63,10 @@ class RestResponseExceptionHandler {
 
     @ExceptionHandler(RemoteFileConnectionRefusedException::class)
     fun handleConnectionRefused(exception: RemoteFileConnectionRefusedException) =
-            makeResponseEntity("Specification couldn't be retrieved from ${exception.location}.", HttpStatus.NOT_FOUND)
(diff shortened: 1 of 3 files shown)


backend/src/main/kotlin/com/tngtech/apicenter/backend/config/RestResponseExceptionHandler.kt
```diff
@@ -63,6 +63,10 @@ class RestResponseExceptionHandler {
 
     @ExceptionHandler(RemoteFileConnectionRefusedException::class)
     fun handleConnectionRefused(exception: RemoteFileConnectionRefusedException) =
-            makeResponseEntity("Specification couldn't be retrieved from ${exception.location}.", HttpStatus.NOT_FOUND)
```
fdw: Why not keep the error message?

catenoid: Is what's written in the PR description ok justification?

fdw: Oh yes, but I should read it :p

PR #176, opened 2019-07-03
Add ID to user defined metadata form

catenoid:
Add ID as a field (like title/version) when inputting metadata for files which don't provide it in their contents (ie. GraphQL). Because of this, previously there was no way (through the frontend) to associate a new version with a pre-existing service.

Also, since deleting the "edit-specification"/pencil button in the overview a while ago, there is no longer a route to `/add-specifications/:id`, hence the body of the `ngInit` is deleted.

Maybe the "edit-specification" button should make a return? Otherwise it is not possible through the frontend to add a new version of an OpenAPI service which doesn't have an `x-api-id`. Or is that a good thing?

### DIFF
--- a/backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceController.kt
+++ b/backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceController.kt
@@ -54,7 +54,7 @@ class ServiceController @Autowired constructor(
     private fun getConsistentId(specificationFileDto: SpecificationFileDto, specificationIdFromPath: String): String {
         specificationFileDto.id?.let {
             idFromFile -> if (idFromFile != specificationIdFromPath)
-                throw MismatchedServiceIdException(idFromFile, specificationIdFromPath)
+                throw MismatchedServiceIdException(listOf(idFromFile, specificationIdFromPath))
         }
         return specificationIdFromPath
     }
--- a/backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataParser.kt
+++ b/backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataParser.kt
@@ -96,7 +96,8 @@ class SpecificationDataParser @Autowired constructor(
                                   metadata: SpecificationFileMetadata? = null
     ): SpecificationMetadata {
         val idFromUpload = extractId(fileContent)
-        val serviceId = getConsistentServiceId(idFromUpload, idFromPath)
+        val serviceId = getConsistentServiceId(listOf(idFromUpload, idFromPath, metadata?.id))
+
         val version = metadata?.version ?: extractVersion(fileContent)
         try {
             SemVer.valueOf(version)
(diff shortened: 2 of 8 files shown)


fdw:
Regarding your question, I don't want to enforce the `x-api-id`. So there should still be a way to upload a new spec for an existing service.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceController.kt
```diff
@@ -54,7 +54,7 @@ class ServiceController @Autowired constructor(
     private fun getConsistentId(specificationFileDto: SpecificationFileDto, specificationIdFromPath: String): String {
         specificationFileDto.id?.let {
             idFromFile -> if (idFromFile != specificationIdFromPath)
-                throw MismatchedServiceIdException(idFromFile, specificationIdFromPath)
+                throw MismatchedServiceIdException(listOf(idFromFile, specificationIdFromPath))
```
fdw: Why not a `Pair`? Then it would be clearer that the two `String`s are related to each other.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataParser.kt
```diff
@@ -96,7 +96,8 @@ class SpecificationDataParser @Autowired constructor(
                                   metadata: SpecificationFileMetadata? = null
     ): SpecificationMetadata {
         val idFromUpload = extractId(fileContent)
-        val serviceId = getConsistentServiceId(idFromUpload, idFromPath)
+        val serviceId = getConsistentServiceId(listOf(idFromUpload, idFromPath, metadata?.id))
```
fdw: So, either that should be a `vararg` where you can put any amount of ids in (and the order doesn't matter), or it should be a specific type like `Triple`.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/service/SpecificationDataParser.kt
```diff
@@ -126,10 +127,12 @@ class SpecificationDataParser @Autowired constructor(
 
     }
 
-    private fun getConsistentServiceId(idFromUpload: String?, idFromPath: String?): ServiceId {
-        if (idFromUpload != null && idFromPath != null && idFromUpload != idFromPath) {
-            throw MismatchedServiceIdException(idFromUpload, idFromPath)
+    private fun getConsistentServiceId(ids: List<String?>): ServiceId {
+        val notNullIds: List<String> = ids.filterNotNull()
```
fdw: What should happen if you manually enter the same id as the `x-api-id`. In my opinion, it should work.

frontend/src/app/specification-form/specification-form.component.html
```diff
@@ -21,14 +21,21 @@ <h4 class="modal-title">Local file upload</h4>
...
            For a GraphQL specification, ApiCenter needs some additional metadata:
          </p>
-         <div *ngFor="let fieldName of objectKeys(additionalFields)">
+         <div *ngFor="let fieldName of objectKeys({
+                                          title: additionalFields.title,
+                                          version: additionalFields.version,
+                                          description: additionalFields.description
+                                          })">
            <input type="text"
                   [placeholder]="fieldName.charAt(0).toUpperCase() + fieldName.slice(1)"
```
fdw: This has nothing to do with this PR, but I think the description should be a `textbox` while `title` and `version` should be a usual `<input type="text">`.

catenoid:
Superseded by #178

Issue #30, opened 2018-10-12
Distinguish "owned" specifications in the list


alexkaserbacher:
Distinguish on specification overview between "general" specifications and specifications created by the logged in user. First of all, there needs to be an association on the data model and within the backend domain. There are some possibilities on how to distinguish your own and other specifications in the frontend:
- Show your own specifications on top of the specification list
- Somehow visually tag your own specification
- Provide two seperate lists of specifications (e.g. "All Specifications" & "Your Specifications") where you can switch between both of them via tabs

fdw:
I think this very closely related to #5: It should be possible to associate a specification with more than one user (i.e. the responsible developers), and it should be possible to have different rights for these users. Maybe we need a relation `specification`-`user`-`permission level`? This would also allow for users to have unusually few permissions.

Or should this "author" association be totally separate from permission management? However, this will probably confuse users.

alexkaserbacher:
Correct. Although, I see this as kind of a "precondition" for permission-management. This issue has solely the purpose of associating users with specifications and generating views for that such that the user can somehow distinguish between "his own" specifications and others.

Permission management would then be an addition to that, such that you can specify what others can see/edit about your specifications.

I thought it would be better to split this into two separate issues.

fdw:
I'm not sure if separating both issues isn't more work, that's why I commented yesterday: If I understand this issue correctly, each spec has exactly one "author" who can then (with #5) change its permissions. 
However, I don't think that one author is the best idea - a spec is usually developed by a team, which must be stored and managed differently. Therefore I think it might make sense to include permission management at least in the design, so that we don't need to rework everything.

In essence, I think we want an ACL for the specs, so I'd say we use that already for this ticket even if we don't need its full might.

alexkaserbacher:
I agree on the implementation details (i.e. instantiating a many-to-many instead of a one-to-many between specifications and users). Interesting question: should we introduce permission-management on versions too? Or is the granularity of specification-permissions already enough!?

Regarding the splitting of the issues: this issue has one focus that does not directly come to mind when thinking about permission management - as I mentioned before, the focus is to visually seperate your "owned" specifications (i.e. later on also specifications that you have write acces to) from all the other specifications, as I may want to quickly see all my owned specifications as a developer (since I'm obligated to maintain them, and/or delete them).

fdw:
[...]

PR #179, opened 2019-07-05
Permissions management

### LINKED ISSUE #5: Permission Management

catenoid:
Close #5 

Demo: (The left window's username is `left`, the right window's username is `right`)

permissions-demo
---
Regarding #30: Should the owner of a service be handled as a special type of permission (granted on upload, potentially transferred)? The logic appears quite similar. Or should the owner be a field in the Service object? Services still have owners, even if we ignored view/edit rights entirely. Or something else? I don't know what would be a natural solution.


fdw:
I'm sorry, I haven't yet had time to look at the code, but at least I can answer question: I'm still not a fan of a separate "owner" permission/role. Rather, I would want to use "maintainer" for this. Obviously, this should automatically be granted to the uploading user, and they can then add other users as maintainers.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/AccessRecordEntity.kt
```diff
@@ -0,0 +1,13 @@
+package com.tngtech.apicenter.backend.connector.database.entity
+
+import org.hibernate.search.annotations.IndexedEmbedded
+import javax.persistence.EmbeddedId
+import javax.persistence.Entity
+
+@Entity
+data class AccessRecordEntity(
+        @EmbeddedId @IndexedEmbedded val accessRecordId: AccessRecordId,
+        val view: Boolean,
```
fdw: Can't we model that as an `enum`, as each one implies the earlier ones? (I.e. `viewPrereleases` needs `view`, and `edit` needs `viewPrereleases`.)

Or can we somehow check these relations when setting them?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/AccessRecordId.kt
```diff
@@ -0,0 +1,10 @@
+package com.tngtech.apicenter.backend.connector.database.entity
+
+import java.io.Serializable
+import javax.persistence.Embeddable
+
+@Embeddable
+data class AccessRecordId(
```
fdw: That isn't possible without a new class?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/AccessRecordDatabase.kt
```diff
@@ -0,0 +1,58 @@
...
+        val viewPrereleases: Boolean = (permission == PermissionType.VIEWPRERELEASE) ||
+            record.map { entity -> entity.viewPrereleases }.orElse(false)
+
+        val edit: Boolean = (permission == PermissionType.EDIT) ||
+            record.map { entity -> entity.edit }.orElse(false)
+
+        accessRecordRepository.save(AccessRecordEntity(key, view, viewPrereleases, edit))
+    }
+
+    override fun removePermission(userId: String, serviceId: ServiceId, permission: PermissionType) {
```
fdw: I'm a bit confused. Do I specify which permissions to remove? Would a generic `setPermissions()` be better?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/AccessRecordDatabase.kt
```diff
@@ -0,0 +1,58 @@
...
+            val view = if (permission == PermissionType.VIEW) false else entity.view
+            val viewPrereleases = if (permission == PermissionType.VIEWPRERELEASE) false else entity.viewPrereleases
+            val edit = if (permission == PermissionType.EDIT) false else entity.edit
+            accessRecordRepository.save(AccessRecordEntity(key, view, viewPrereleases, edit))
+        }
+    }
+
+    override fun hasPermission(userId: String, serviceId: ServiceId, permission: PermissionType): Boolean {
+        val key = AccessRecordId(serviceId.id, userId)
+        val record = accessRecordRepository.findById(key)
```
fdw: Do we cache those results automatically?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceController.kt
```diff
@@ -52,6 +54,28 @@ class ServiceController @Autowired constructor(
         return specificationFileDtoMapper.fromDomain(specification)
     }
 
+    @PutMapping("/{serviceId}/chmod/{userId}")
+    fun chmodService(@PathVariable serviceId: String,
```
fdw: Can we please rename that? "chmod" is not a verb ;)

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceController.kt
```diff
@@ -52,6 +54,28 @@ class ServiceController @Autowired constructor(
...
+                     @RequestParam(value = "edit", defaultValue = "false") edit: String
+    ) {
+        val id = ServiceId(serviceId)
+        serviceHandler.changePermission(id, userId, view.toBoolean(), PermissionType.VIEW)
+        serviceHandler.changePermission(id, userId, viewPrereleases.toBoolean(), PermissionType.VIEWPRERELEASE)
+        serviceHandler.changePermission(id, userId, edit.toBoolean(), PermissionType.EDIT)
+    }
+
+    @GetMapping("/{serviceId}/getmod/{userId}")
+    fun getmodService(@PathVariable serviceId: String,
```
fdw: Why do we need this? The controller should already check whether the current user has the necessary permissions before it returns the spec.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/security/JwtAuthenticationProvider.kt
```diff
@@ -24,5 +25,20 @@ class JwtAuthenticationProvider : AuthenticationProvider {
...
+                .build()
+                .verify(authorizationHeader.replace("Bearer ", ""))
+                .subject
+    }
+
+    fun getCurrentUser() =
+            try {
+                (SecurityContextHolder.getContext().authentication as JwtAuthenticationToken).userId
+            } catch (exception: ClassCastException) {
+                // Thrown when running integration tests
```
fdw: Why? Can't we adapt our integration tests?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/ServiceController.kt
catenoid: This is only used in the edit-permissions frontend form, to pre-populate the permissions of a certain user. It is not used for authorization anywhere (see the demo gif)

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/AccessRecordId.kt
catenoid: My interpretation of the type parametrisation of a repository was that it requires an ID class? eg. `interface MyRepository<T, ID extends Serializable>
  extends JpaRepository<T, ID>`

fdw: That's unfortunate, but c'est la vie :p

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/handler/ServiceHandler.kt
```diff
@@ -1,28 +1,46 @@
...
 ) {
 
-    fun addNewSpecification(specification: Specification, serviceId: ServiceId, fileUrl: String?) {
+    fun addNewSpecification(specification: Specification, serviceId: ServiceId, fileUrl: String?, isPublic: Boolean) {
         val service: Service? = servicePersistor.findOne(serviceId)
 
         if (service == null) {
             saveNewService(specification, serviceId, fileUrl)
-        } else {
+            val userId = jwtAuthenticationProvider.getCurrentUserId()
```
fdw: How threadsafe is that?

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/handler/ServiceHandler.kt
```diff
@@ -61,14 +79,68 @@ class ServiceHandler @Autowired constructor(
 
     }
 
-    fun findAll(pageNumber: Int, pageSize: Int): ResultPage<Service> =
-            servicePersistor.findAll(pageNumber, pageSize)
+    fun findAll(pageNumber: Int, pageSize: Int): ResultPage<Service> {
+        val userId = jwtAuthenticationProvider.getCurrentUserId()
+        val page = servicePersistor.findAll(pageNumber, pageSize, userId, apiCenterProperties.getAnonymousUsername())
+        return ResultPage(this.filterByViewPermission(page.content), page.last)
```
fdw: I could imagine either something like a "findAllWithPermissions" method or a `filter` method on `Page`.

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/handler/ServiceHandler.kt
```diff
@@ -1,28 +1,46 @@
...
+import com.tngtech.apicenter.backend.domain.exceptions.PermissionDeniedException
 import com.tngtech.apicenter.backend.domain.exceptions.SpecificationConflictException
 import com.tngtech.apicenter.backend.domain.exceptions.SpecificationDuplicationException
+import com.tngtech.apicenter.backend.domain.service.PermissionsManager
 import com.tngtech.apicenter.backend.domain.service.ServicePersistor
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
 @Component
 class ServiceHandler @Autowired constructor(
```
fdw: I'm wondering if it would be possible to have something like the previous `ServiceHandler` (without permissions), and a decorator for that that takes care of the permissions stuff.

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/handler/SessionHandler.kt
```diff
@@ -13,23 +13,14 @@ class SessionHandler constructor(
 
     fun authenticate(username: String, password: String): User? {
         val searchUser = externalAuthenticator.authenticate(username, password) ?: return null
+        // A new uuid is assigned on every authentication
```
fdw: That comment makes no sense. Where does the UUID come from?

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/handler/SpecificationHandler.kt
```diff
@@ -0,0 +1,48 @@
...
+        private val specificationPersistor: SpecificationPersistor,
+        private val jwtAuthenticationProvider: JwtAuthenticationProvider,
+        private val permissionsManager: PermissionsManager
+) {
+    fun findOne(serviceId: ServiceId, version: String): Specification? {
+        val userId = jwtAuthenticationProvider.getCurrentUserId()
+
+        val specification = specificationPersistor.findOne(serviceId, version)
+        val canView = when (specification?.metadata?.releaseType) {
+            ReleaseType.PRERELEASE -> permissionsManager.hasPermission(userId, serviceId, PermissionType.VIEWPRERELEASE)
```
fdw: I'm unsure whether to push this logic into `ReleaseType`. Thoughts?

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/service/ServicePersistor.kt
```diff
@@ -3,10 +3,11 @@ package com.tngtech.apicenter.backend.domain.service
 import com.tngtech.apicenter.backend.domain.entity.ResultPage
 import com.tngtech.apicenter.backend.domain.entity.ServiceId
 import com.tngtech.apicenter.backend.domain.entity.Service
+import java.util.*
 
 interface ServicePersistor {
     fun save(service: Service)
-    fun findAll(pageNumber: Int, pageSize: Int): ResultPage<Service>
+    fun findAll(pageNumber: Int, pageSize: Int, username: String, anonymousUsername: String): ResultPage<Service>
```
fdw: So far, I've seen `jwtAuthenticationProvider.getCurrentUserId()` to get the current user id, a `User` object and now just `username` as a parameter. Could we please unify that? Maybe all methods take a `User` parameter, or all get the user from `jwtAuthenticationProvider`.

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/service/ServicePersistor.kt
```diff
@@ -3,10 +3,11 @@ package com.tngtech.apicenter.backend.domain.service
 import com.tngtech.apicenter.backend.domain.entity.ResultPage
 import com.tngtech.apicenter.backend.domain.entity.ServiceId
 import com.tngtech.apicenter.backend.domain.entity.Service
+import java.util.*
```
fdw: That's not needed.

backend/src/test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerUnitTest.kt
```diff
@@ -6,17 +6,16 @@ import com.nhaarman.mockitokotlin2.verify
...
 import com.tngtech.apicenter.backend.connector.rest.mapper.SpecificationFileDtoMapper
 import com.tngtech.apicenter.backend.domain.entity.*
-import com.tngtech.apicenter.backend.domain.service.SpecificationPersistor
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
 
 internal class SpecificationControllerUnitTest {
 
-    private val specificationPersistor: SpecificationPersistor = mock()
+    private val specificationHandler: com.tngtech.apicenter.backend.domain.handler.SpecificationHandler = mock()
```
fdw: Why the fully qualified import? Do we have more than one `SpecificationHandler` class?

frontend/src/app/permissions-form/permissions-form.component.html
```diff
@@ -0,0 +1,33 @@
...
+  <div class="row top-buffer">
+    <div class="alert alert-danger" role="alert" *ngIf="error">
+      {{ error }}
+    </div>
+  </div>
+  <div class="row top-buffer">
+    <div class="col">
+      <h4>Edit permissions</h4>
+
+      <label for="username">Target user:</label>
```
fdw: Just "user" should be enough, I think.

frontend/src/app/permissions-form/permissions-form.component.html
```diff
@@ -0,0 +1,33 @@
...
+      <br>
+      <input type="text" id="username" [(ngModel)]="targetUser" (ngModelChange)="selectorOnChange()"/>
+
+      <br>
+      <br>
+
+      <label for="role">Role to assign:</label>
```
fdw: "New role"?

frontend/src/app/specification-form/specification-form.component.html
```diff
@@ -10,33 +10,41 @@ <h4 class="modal-title">Local file upload</h4>
...
-         </div>
-         <hr>
-         <p>Test API endpoint, where sample requests can be sent:</p>
-         <input type="text" placeholder="URL" [(ngModel)]="additionalFields.endpointUrl"/>
-       </div>
+      <hr>
+
+      <input type="checkbox" id="isPublic" [(ngModel)]="isPublic">
+      &nbsp;
+      <label for="isPublic">Make public</label>
```
fdw: "publicly visible"

frontend/src/app/specification-form/specification-form.component.ts
```diff
@@ -42,6 +42,7 @@ export class SpecificationFormComponent implements OnInit {
   remoteFileUrl: string;
   additionalFields = {title: '', version: '', description: '', endpointUrl: ''};
   id?: string;
+  isPublic: boolean;
   showAdditionalMetadataFields = false;
   objectKeys = Object.keys;
```
fdw: Is that needed?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/AccessRecordEntity.kt
```diff
@@ -0,0 +1,28 @@
...
+        @JoinColumn(name = "username")
+        val userEntity: UserEntity,
+
+        val view: Boolean,
+        val viewPrereleases: Boolean,
+        val edit: Boolean
+) {
+        override fun hashCode(): Int {
+            // Prevent the recursive calls on the ServiceEntity <-> AccessRecordEntity relationship from causing a stack overflow
+                return 0
```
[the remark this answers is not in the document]
catenoid: Hmm... were you looking at the latest version of this PR? I though this way quite an old change, when I first realised about the recursive hashCode calls.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/AccessRecordEntity.kt
```diff
@@ -0,0 +1,28 @@
+package com.tngtech.apicenter.backend.connector.database.entity
+
+import org.hibernate.search.annotations.IndexedEmbedded
+import javax.persistence.*
+
+@Entity
+class AccessRecordEntity(
+        @EmbeddedId @IndexedEmbedded val accessRecordId: AccessRecordId,
+
+        @ManyToOne
```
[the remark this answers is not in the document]
catenoid: As far as I understand, yes, because the parent (ServiceEntity) needs a reference so the remove cascade can remove AccessRecordEntities when the ServiceEntity is deleted. The child needs a reference because the ServiceId is part of it's primary key.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/UserEntity.kt
```diff
@@ -7,8 +7,6 @@ import javax.persistence.Id
...
 @Entity
 data class UserEntity(
-    @Id @GeneratedValue val id: UUID,
-    val username: String,
-    val email: String,
-    val origin: String,
-    val externalId: String)
\ No newline at end of file
+    @Id val username: String,
+    val email: String
```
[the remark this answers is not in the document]
catenoid: Technically no... I suppose I gutted so many fields out of this class, I felt merciful :) Would you like me to remove it? Eventually the email will probably be used.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/mapper/converter/ServiceEntityConverter.kt
```diff
@@ -0,0 +1,26 @@
...
+class ServiceEntityConverter : BidirectionalConverter<Service, ServiceEntity>() {
+
+    override fun convertFrom(source: ServiceEntity, destinationType: Type<Service>?, mappingContext: MappingContext?): Service {
+        val specifications = source.specifications.map { spec -> mapperFacade.map(spec, Specification::class.java) }
+        return Service(id = ServiceId(source.id), description = source.description, remoteAddress = source.remoteAddress, specifications = specifications, title = source.title)
+    }
+
+    override fun convertTo(source: Service, destinationType: Type<ServiceEntity>?, mappingContext: MappingContext?): ServiceEntity {
+        val specifications = source.specifications.map { spec -> mapperFacade.map(spec, SpecificationEntity::class.java) }
+        return ServiceEntity(id = source.id.id, title = source.title, specifications = specifications, remoteAddress = source.remoteAddress, description = source.description, accessRecords = setOf())
```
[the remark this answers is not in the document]
catenoid: No: https://stackoverflow.com/a/39875431

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/repository/ServiceRepository.kt
```diff
@@ -11,4 +16,7 @@ interface ServiceRepository : PagingAndSortingRepository<ServiceEntity, String>
 
     override fun existsById(id: String): Boolean
     override fun findById(id: String): Optional<ServiceEntity>
+
+    @Query("select s from ServiceEntity s inner join AccessRecordEntity a on a.accessRecordId.serviceId = s.id where a.accessRecordId.username = :username")
```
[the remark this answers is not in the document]
catenoid: As a response to this comment, and the few above about `AccessRecordEntity a`; These are JPQL queries by default, which have a slightly different format, eg. `SELECT u FROM User u WHERE u.status = 1` https://www.baeldung.com/spring-data-jpa-query

backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
```diff
@@ -25,19 +23,29 @@ class SpecificationControllerIntegrationTest {
...
+            .with(userRequestPostProcessor)
             .with(csrf()))
             .andExpect(jsonPath("$.metadata.version", equalTo("1.0.0")))
     }
 
+    @Test
+    fun findOneSpecification_requiresAuthentication() {
+        mockMvc.perform(get("/api/v1/service/b6b06513-d259-4faf-b34b-a216b3daad6a/version/1.0.0")
+                .with(csrf()))
+                .andExpect(status().isNotFound)
```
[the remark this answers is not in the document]
catenoid: We did, I think quite a bit :) What I remember is a policy of not leaking information about what services exist, if a user doesn't have view permission. Failed views should get 404 instead of anything indicating the operation failure comes from insufficient permission. Am I mistaken?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/ServiceEntity.kt
```diff
@@ -15,5 +15,8 @@ data class ServiceEntity(
         mappedBy = "service",
         cascade = [CascadeType.ALL]
     ) @OrderBy("created DESC") val specifications: List<SpecificationEntity>,
```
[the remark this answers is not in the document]
catenoid: To confuse you, of course 😄 (I didn't write these lines, but will change them...)

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/repository/AccessRecordRepository.kt
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.data.repository.query.Param
+
+interface AccessRecordRepository: CrudRepository<AccessRecordEntity, AccessRecordId> {
+    @Query("select count(a)>0 from AccessRecordEntity a where a.role = 2 and a.accessRecordId.serviceId = :serviceId and a.accessRecordId.username <> :username")
+    fun otherEditorsExist(@Param("serviceId") serviceId: String,
+                          @Param("username") userToDowngrade: String): Boolean
+
+    @Modifying
+    @Query("delete from AccessRecordEntity a where a.accessRecordId.serviceId = :serviceId")
+    fun clearPermissions(@Param("serviceId") serviceId: String)
```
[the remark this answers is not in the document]
catenoid: My mistake, `clearPermissions` was obsolete ever since the remove cascade on the ServiceEntity -> AccessRecordEntity worked. Nothing in the PR actually uses this function, I'll take it out.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/AccessRecordDatabase.kt
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private fun checkRole(permissionType: PermissionType, role: Role): Boolean {
+        val viewer = EnumSet.of(PermissionType.VIEW)
+        val viewerX = EnumSet.of(PermissionType.VIEW, PermissionType.VIEWPRERELEASE)
+        val editor = EnumSet.of(PermissionType.VIEW, PermissionType.VIEWPRERELEASE, PermissionType.EDIT)
+
+        return when (role) {
+            Role.VIEWER -> viewer.contains(permissionType)
+            Role.VIEWER_X -> viewerX.contains(permissionType)
```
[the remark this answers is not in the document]
catenoid: Sorry, I'm not convinced yet :) Viewer_X to me is an 'augmented viewer'. I find viewerPre less descriptive, even a bit confusing. And the augmented viewers may not actually be Testers? 🤷‍♂

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/ApiCenterProperties.kt
```diff
@@ -5,8 +5,10 @@ import org.springframework.boot.context.properties.ConfigurationProperties
 @ConfigurationProperties("apicenter")
 class ApiCenterProperties{
     private var pageSize = 10
+    private var anonymousUsername = "public"
 
     fun getPageSize(): Int = pageSize
+    fun getAnonymousUsername(): String = anonymousUsername
```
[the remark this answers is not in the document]
catenoid: I have no better answer than I followed the example here https://www.baeldung.com/configuration-properties-in-spring-boot for which I assume the reasoning is you may want to check properties before setting them, and so having explicit getters and setters for everything makes the class style-wise more consistent.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/UserDatabase.kt
```diff
@@ -1,23 +1,33 @@
...
         userRepository.save(userEntityMapper.fromDomain(user))
     }
 
-    override fun findByOrigin(origin: String, externalId: String) =
-        userRepository.findByOriginAndExternalId(origin, externalId)
+    override fun findById(username: String): User? =
+            userRepository.findById(username).orElse(null)?.let { spec -> userEntityMapper.toDomain(spec) }
 
-    override fun exists(origin: String, externalId: String) = userRepository.checkExistenceByOrigin(origin, externalId)
+    @Transactional(isolation = Isolation.SERIALIZABLE)
```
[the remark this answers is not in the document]
catenoid: This was in response to the following scenario: User logs in for the first time. The `login` button bounces, and the session request is sent twice. ApiCenter checks for the existence of this user, and when there is no Transactional annotation, returns false for both requests, and so tries to create two UserEntities. With the annotation, this doesn't happen.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/security/JwtAuthenticationProvider.kt
```diff
@@ -2,27 +2,38 @@ package com.tngtech.apicenter.backend.connector.rest.security
...
         val token = (authentication as JwtAuthenticationToken).token
 
-        val user = JWT.require(Algorithm.HMAC512(jwtSecuritySecret.toByteArray()))
+        val username = JWT.require(Algorithm.HMAC512(jwtSecuritySecret.toByteArray()))
             .build()
             .verify(token)
             .subject
 
-        return if (user != null) JwtAuthenticationToken(user, token) else null
+        return if (username != null) JwtAuthenticationToken(username, token) else null
```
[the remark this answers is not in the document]
catenoid: Perhaps, but I don't really think it's part of this PR, and we're already nearing 900 lines ;)

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/handler/SessionHandler.kt
catenoid: True, it is an artefact of when a User's ID was a UUID, and I forgot to change it. (Now I see why people say "code never lies, comments sometimes do")

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/handler/ServiceHandler.kt
catenoid: This pre-release filter function is used in other places in the handler, for non-paged requests as well, so I would like to keep it in this class. And I didn't want to put this logic in the servicePersistor, because it is permissions logic, and not strictly the responsibility of a storage service. (Of course, I had to compromise on that a bit, because the paging request query needs to know which user is logged in, so that the number of services per page is consistent regardless of who is logged in... 🤷‍♂)

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/handler/ServiceHandler.kt
catenoid: Sorry, I don't understand the question... what is the danger?

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/service/ServicePersistor.kt
catenoid: [Firstly, my mistake, anything called userId is now username, as that is now the id of a user. The Id used to be an auto-generated UUID. I've renamed `userId` where it appeared.] Secondly, I see your point, about consistently representing the idea of a User in the code. However there are sortof two representations of a user: who claims to be authenticated (handled by the token verification, a property of the session) and additional info attached to a certain user (handled by the UserPersistor / User object, which for now only includes an email address). I could perform a findById(), which links the two concepts by the username, but for permissions management, we fundamentally do not need any other information than the (verified) username claim. Thoughts?

catenoid:
> 
> 
> Okay, that was a long one...

When it rains, it pours, as we say 😉

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/AccessRecordDatabase.kt
catenoid: Sorry I don't understand, can you explain?

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/entity/Role.kt
```diff
@@ -0,0 +1,22 @@
...
+        override fun permissions(): EnumSet<PermissionType> =
+                EnumSet.of(PermissionType.VIEW, PermissionType.VIEWPRERELEASE)
+    },
+
+    EDITOR {
+        override fun permissions(): EnumSet<PermissionType> =
+                EnumSet.of(PermissionType.VIEW, PermissionType.VIEWPRERELEASE, PermissionType.EDIT)
+    };
+
+    abstract fun permissions(): EnumSet<PermissionType>
```
fdw: Instead of a function, you can define fields for enums:
```
enum class Role(permissions: EnumSety<PermissionType> {
    VIEWER(EnumSet.of(VIEW))
    VIEWER_X(EnumSet.of(VIEW, VIEW_PRERELEASE)
...
```

backend/src/integration-test/kotlin/com/tngtech/apicenter/backend/connector/rest/controller/SpecificationControllerIntegrationTest.kt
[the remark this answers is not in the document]
fdw: Yes, with that explanation, I still agree :)
I was just wondering if this was the occassion for the discussion, or something different ;)

backend/src/main/kotlin/com/tngtech/apicenter/backend/config/ApiCenterProperties.kt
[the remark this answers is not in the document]
fdw: I'd rather have the usual `data class` stuff, if possible. It makes everything more consistent.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/AccessRecordEntity.kt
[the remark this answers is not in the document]
fdw: Okay, that (unfortunately) makes sense. Thanks!

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/entity/UserEntity.kt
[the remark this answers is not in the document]
fdw: Okay, show mercy. This time :P

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/mapper/converter/ServiceEntityConverter.kt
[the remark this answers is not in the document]
fdw: :(

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/AccessRecordDatabase.kt
fdw: Does every one of these method calls end up as a call to the database, or are they cached in between to relieve the DB?

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/UserDatabase.kt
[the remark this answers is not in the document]
fdw: :+1:

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/rest/security/JwtAuthenticationProvider.kt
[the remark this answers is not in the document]
fdw: So? :p 
No, I agree, let's do it separately

backend/src/main/kotlin/com/tngtech/apicenter/backend/domain/service/ServicePersistor.kt
fdw: For now, that makes sense. In general, I could see a `User` class that contains the verified `username` and optionally other stuff as well.

backend/src/main/kotlin/com/tngtech/apicenter/backend/connector/database/service/AccessRecordDatabase.kt
catenoid: I don't think any caching is currently enabled https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/cache/annotation/EnableCaching.html

catenoid:
The failing test is `findAllSpecifications`; I believe since adding the `distinct` keyword into the query, the order of the results is variable. [I added distinct because one `ServiceEntity` can now have potentially two `AccessRecordEntities` applying to the same user (one for the user, one for the public/anonymous username), and otherwise a service will appear twice as a result of the join. Maybe there is a better solution?]

fdw:
So it's rather by chance that the test worked beforehand? ;)

I think an `order by name` is in order.

PR #190, opened 2019-07-08
Make description field a textarea

catenoid:
Also uses `<label>` elements instead of the placeholder text for describing form fields; semantic html is recommended for accessibility / screen readers.

Currently the metadata form looks like:

form-example

### DIFF
--- a/frontend/src/app/specification-form/specification-form.component.html
+++ b/frontend/src/app/specification-form/specification-form.component.html
@@ -22,20 +22,38 @@
        <div *ngIf="showAdditionalMetadataFields" [@displayFields]>
          <hr>
          <p>
-           For a GraphQL specification, ApiCenter needs some additional metadata:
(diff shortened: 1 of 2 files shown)


frontend/src/app/specification-form/specification-form.component.html
```diff
@@ -22,20 +22,38 @@ <h4 class="modal-title">Local file upload</h4>
...
-                                          version: additionalFields.version,
-                                          description: additionalFields.description
-                                          })">
-           <input type="text"
-                  [placeholder]="fieldName.charAt(0).toUpperCase() + fieldName.slice(1)"
-                  [(ngModel)]="additionalFields[fieldName]"/>
-         </div>
+
+         <label for="title-field">Title</label>
+         &nbsp;
```
fdw: Why do we need that `&nbsp`?

PR #204, opened 2019-07-26
Add more error cases to diff view

catenoid:
@fdw Do you have any advice on the TravisCI errors?

### DIFF
--- a/frontend/src/app/diff-versions/diff-versions.component.html
+++ b/frontend/src/app/diff-versions/diff-versions.component.html
@@ -106,14 +106,21 @@
           </ngb-tab>
         </ngb-tabset>
       </div>
-      <div *ngIf="!this.validRangeSelected">
-        <p>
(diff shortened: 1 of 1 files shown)


frontend/src/app/diff-versions/diff-versions.component.html
```diff
@@ -106,14 +106,21 @@ <h5>Endpoint changes</h5>
           </ngb-tab>
         </ngb-tabset>
       </div>
-      <div *ngIf="!this.validRangeSelected">
-        <p>
-          Choose two different versions from the table on the left using the radio buttons
-        </p>
+      <div *ngIf="searchComplete ? this.service.specifications.length >= 2 : false">
```
fdw: Can't you specify that using `&&` rather than with a nested if? ;)

catenoid: It would make the conditional expressions really long, no? Is it necessarily better?

PR #406, opened 2020-06-06
GitHub Actions: use checkout v2

sullis:


fdw:
Hi @sullis , thanks for the PR! It seems like it has some problems - do you know what changed between v1 and v2 that could trigger this?

sullis:
> Hi @sullis , thanks for the PR! It seems like it has some problems - do you know what changed between v1 and v2 that could trigger this?

I am not aware of anything. I am using checkout@v2 on all of my libraries.

fdw:
We're working on a rewrite, you can see the current progress on the branch architecture-inventory. There, checkout is already updated to v2.

If this is good enough for you, I would close this PR.

PR #497, opened 2020-10-02
Mixed Updates

fdw:
* Update backend dependencies
* Update Gradle wrapper
* Update to Java 11
* Make Docker images from `bootBuildImage` work


alexkaserbacher (changes requested):
There seems to be a problem when executing the E2E-Tests. More specifically the task `:backend:bootRun` seems to not work in the CI pipeline.

fdw:
There is one failing test in `:backend:test`, that also fails on `main`.

The thing with the cypress tests is that I upgraded to Java 11, but cypress, for some reason, still starts it on a JVM 8. Do you have an idea where I can tell cypress/gradle the correct Java version?

PR #540, opened 2020-12-23
Fix DB relations

fdw:
In short, the `OneToMany` was set incorrectly, which is subtle at the best of times, and was downright evil here.
See the commit message for a longer description.


alexkaserbacher:
Nice catch, thank you very much :)

alexkaserbacher (approved):

PR #657, opened 2021-04-18
ci: trigger workflow on pull_request

sullis:


fdw:
Hey, first off: Thank you!

However, it was a conscious decision to let it run on all branches and not just `main`.

Regarding pull requests, shouldn't that also be covered by `push`? Previously, pushing and opening a PR would start all actions twice, which is not ideal. Is that different now? And is there a way to open a PR without also pushing?

sullis:
> 
> Regarding pull requests, shouldn't that also be covered by `push`? 

Unfortunately no.  If you examine https://github.com/TNG/ApiCenter/pull/655 you'll notice that the PR did not trigger any GitHub Action workflows.

fdw:
Hm. It started one check (DCO), but the others do have some problem. However, this one had all checks run. Did you do something different here?

